PR #1, opened 2022-08-29
legacy support: Cherenkov data model for juggler `IRTAlgorithm`

c-dilks:
#### Do not merge

Supports Juggler IRT usage for the dRICH.

Moved from https://eicweb.phy.anl.gov/EIC/eicd/-/merge_requests/70


c-dilks:
We've been testing this with `irt` in `juggler` for a while, and it seems to be good enough for now. We can revise this as we rewrite things in `EICrecon`.

c-dilks:
This is the last thing needed for the initial PR for integrating IRT into EICrecon; see https://github.com/eic/EICrecon/issues/255

wdconinc (commented):
I have a hard time seeing the anticipated data flow here:
1. You start from hits on pixels, which are currently tracker hits, and which presumably we continue to treat with tracker hit digitization and tracker hit reconstruction.
2. What is next then? Do you create a `CherenkovThetaAngleMeasurement` from the reconstructed hits? What would be a measurement outcome, without model independence, but `CherenkovThetaAngleMeasurement` contains model output like radiator, npe, etc.
3. How does this relate to the hypothesis? It seems you fill the hypothesis at the same time as the measurements, which seems to indicate a tight connection between your 'measurement' and 'hypothesis'.

Anyway, can you explain that a bit better?

In particular I'm worried about how this will work for pfrich and/or mrich? If it is a data model, it needs to be generic enough to allow for (in particular) the mrich to use this as well. Was there any discussion with Murad?

c-dilks:
Let's revisit this later when it's more clear what we need. We may be able to use just what is available in `EDM4hep`. Until we have working IRT in `EICrecon`, let's leave this PR in draft state since it supports the [Juggler IRT code](https://eicweb.phy.anl.gov/EIC/juggler/-/blob/73-add-rich-irt-algorithm/JugPID/src/components/IRTAlgorithm.cpp)

c-dilks:
Closing this PR, but please do not delete the branch (in case @chchatte92 needs it)

PR #6, opened 2022-09-04
fix: add EDM4hep include dir

wdconinc:
This addresses an issue when EDM4hep is installed in a dedicated prefix and the include directory does not end up in the compilation include line by virtue of other includes (as is the case when everything is in /usr/local).

New include directory in the top level CMakeLists.

- [X] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, in that it makes the default compilation succeed on some systems where it previously didn't.


wdconinc:
See also discussion at https://github.com/eic/eicd/pull/6.

veprbl (approved):

PR #5, opened 2022-09-03
cmake: explicitly link to EDM4hep::edm4hep

veprbl:
### Briefly, what does this PR introduce?

https://github.com/AIDASoft/podio/blob/a8efda986f776892e90e7aafa4958aaf18a1e6c7/doc/datamodel_syntax.md?plain=1#L157

Fixes linker errors on macOS:
```
Undefined symbols for architecture x86_64:
  "edm4hep::MCParticle::MCParticle(edm4hep::MCParticleObj*)", referenced from:
      edm4eic::MCRecoParticleAssociation::getSim() const in MCRecoParticleAssociation.cc.o
      edm4eic::operator<<(std::__1::basic_ostream<char, std::__1::char_traits<char> >&, edm4eic::MCRecoParticleAssociation const&) in MCRecoPart
icleAssociation.cc.o
      edm4eic::MutableMCRecoParticleAssociation::getSim() const in MutableMCRecoParticleAssociation.cc.o
      edm4eic::MCRecoClusterParticleAssociation::getSim() const in MCRecoClusterParticleAssociation.cc.o
      edm4eic::operator<<(std::__1::basic_ostream<char, std::__1::char_traits<char> >&, edm4eic::MCRecoClusterParticleAssociation const&) in MCR
ecoClusterParticleAssociation.cc.o
      edm4eic::MutableMCRecoClusterParticleAssociation::getSim() const in MutableMCRecoClusterParticleAssociation.cc.o
      edm4eic::MCRecoTrackParticleAssociation::getSim() const in MCRecoTrackParticleAssociation.cc.o
      ...
```

### What kind of change does this PR introduce?
- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?

Linking to more libraries should be generally safe. No additional changes are needed.

### Does this PR change default behavior?

No.

PR #11, opened 2022-09-19
fix: no need to `target_include_directories`

wdconinc:
This removes the `target_include_directories` introduced in #6 again, since the `target_link_libraries` introduced in #5 set the include directories the right way.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: technical debt reduction

No.

No.


veprbl (approved):
Seems like this should work, since the headers should, indeed, be appended to the target:

https://github.com/key4hep/EDM4hep/blob/50759186bc72489a2c2731414affd7624606115d/edm4hep/CMakeLists.txt#L6
and
https://github.com/AIDASoft/podio/blob/e1096884df7c42e44ed3c51640f2d1cab110bab5/cmake/podioMacros.cmake#L238-L240

PR #17, opened 2022-12-18
feat: add `CherenkovParticleID` datatype

c-dilks:
Add new `datatype` for RICH PID: `CherenkovParticleID`. While we intend to use `edm4hep::ParticleID` for the final RICH PID, it is useful to have an intermediate, more detailed datatype specific for Cherenkov PID.

This is a refactor of what we used in Juggler for ATHENA; the branch [`irt-data-model`](https://github.com/eic/EDM4eic/tree/irt-data-model) should remain for legacy Juggler support (see #1) until we are fully transitioned to EICrecon.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators: @chchatte92 @alexander-kiselev 

No

No


c-dilks:
It does not seem possible to create 1-N relations on the fly from `CherenkovParticleID` -> `edm4hep::ParticleID` in such a way that `EICrecon` does not corrupt. Instead we revert to a vector of `edm4eic::CherenkovPdgHypothesis` custom components, as in #1.

c-dilks:
This is ready for review, but with a few comments/questions:
- Should `wavelength` units be in `nm` (somewhat preferred) or conform to other lengths and be in `mm`?
- See above comment about "on the fly 1-N relations" `CherenkovParticleID` -> `edm4hep::ParticleID`; I don't think it's worth the effort to try this again until PODIO support in EICrecon/JANA is improved
- Ideally all Cherenkov PID subsystems can use this `CherenkovParticleID` datatype. I tried to solicit feedback from other ePIC Cherenkov PID subsystem experts, but received none; therefore, we just let them make their own modification requests when they are ready

wdconinc (commented):
Can you mock up an event with this filled? I'm wondering about the `npe` field in both CerenkovParticleID` and `CerenkovPdgHypothesis`. If you observe, say, 12 pe in an event it goes in `CerenkovParticleID`, and if 11 are predicted for an electron, that goes in the hypothesis, I assume. What about covariance on `npe` for that hypothesis? Presumably you need it for the likelihood.

wdconinc:
> Should wavelength units be in nm (somewhat preferred) or conform to other lengths and be in mm?

mm only preferred as a distance measure. Units should be chosen so they don't lead to internal inconsistencies in likely scenarios when doing arithmetic with related quantities. Here the inconsistencies would be due to hit energy deposition which for a standard tracking detector is stored in GeV. So, I'd almost suggest to store the photon energy in GeV, which I know is silly but I know is what geant4 does (well, MeV is their preferred unit, but that's what an optical photon sensitive detector stores).

edm4eic.yaml
```diff
@@ -140,6 +140,13 @@ components:
       - edm4eic::Cov2f    directionError  // Error on the polar and azimuthal angles
       - float             pathlength      // Pathlength from the origin to this point
       - float             pathlengthError // Error on the pathlenght
+
+  ## PID hypothesis from Cherenkov detectors
+  edm4eic::CherenkovPdgHypothesis:
```
wdconinc, line 145: Not too keen on the `Pdg` which doesn't seem like consistent capitalization for an acronym. It's also not really only PDG anymore, it's any PID code.

How about
```suggestion
  edm4eic::CherenkovPIDHypothesis:
```

Or
```suggestion
  edm4eic::CherenkovParticleIDHypothesis:
```

edm4eic.yaml
```diff
@@ -275,8 +282,23 @@ datatypes:
       - int32_t           sector            // The sector this hit occured in
       - edm4hep::Vector3f local             // The local position of the hit in detector coordinates (relative to the sector) [mm]
 
+  edm4eic::CherenkovParticleID:
+    Description: "Cherenkov detector PID"
+    Author: "A. Kiselev, C. Chatterjee, C. Dilks"
+    Members:
+      - int32_t           radiator          // Radiator number
+      - float             theta             // Estimated average Cherenkov cone angle
```
wdconinc, line 290: This is observed, not estimated, right? If estimated, doesn't it depend on the particle type hypothesis?

Isn't this just avg of the vector members and therefore redundant?

edm4eic.yaml
```diff
@@ -275,8 +282,23 @@ datatypes:
       - int32_t           sector            // The sector this hit occured in
       - edm4hep::Vector3f local             // The local position of the hit in detector coordinates (relative to the sector) [mm]
 
+  edm4eic::CherenkovParticleID:
+    Description: "Cherenkov detector PID"
+    Author: "A. Kiselev, C. Chatterjee, C. Dilks"
+    Members:
+      - int32_t           radiator          // Radiator number
```
wdconinc, line 289: Unsigned? Presumably part of cellID?

edm4eic.yaml
```diff
@@ -275,8 +282,23 @@ datatypes:
       - int32_t           sector            // The sector this hit occured in
       - edm4hep::Vector3f local             // The local position of the hit in detector coordinates (relative to the sector) [mm]
 
+  edm4eic::CherenkovParticleID:
```
wdconinc, line 285: Does this need to include a set of PMT hits (tracker hits) that were used? I assume you may decide some were noise. Is there no vector member link to the hits needed?

alexander-kiselev:
Hi Wouter,

  we may want to rework the data model, adjusting it to the emerging algorithmic part. In what I have under the hood for pfRICH now, strictly speaking, there is now covariance on the Npe number. I can see it can be introduced as a free parameter, but as long as these Lagrange multiplier like variables are not needed, there is no point to add them to the problem. Npe enters the chi^2 clause as a separate term in a way there is a penalty if the observed number of hits deviates from the expected Poissonic mean for a given combination of {mass, momentum}. And there is a separate term, which governs the expected count of the background hits in a similar way (this part is commented out now, no time to restore). So, contrary to a rather lousy ATHENA's IRT add-on, for ePIC pfRICH I pretty much realized the algorithm outlined back then on one of the git threads (actually would be nice to find the text one more time), with the added benefit that timing of the aerogel photons is now intrinsically merged into ring imaging, at least conceptually. What is still critically missing is timing from the "main" bulk of photons, produced in the sensor window. But this is also coming. We can talk about the details on Monday.

  Cheers,
    Alexander.


________________________________

alexander-kiselev:
Hi Wouter,

  I think all this old ATHENA crap needs to go, soon. It is not used even in the pfRICH application any longer, where it was introduced for a single reason, essentially to help handle non-gaussian parent distribution of the emission point. Now that we switched to a different type of photosensors and aerogel (much into the UV), the additional single photon angular smearing is so high, that the residuals are pretty much "gaussian" and the task allows a relatively clear chi^2 formulation without resorting to summing up a bunch of uniform distributions.

  Besides this, the PID evaluation is now event-level per design, and the format seemingly does not fit the engine output any longer anyway.

  Chris, once the March rush is over, we should probably make an effort to attach these codes to either just dRICH or dRICH+pfRICH, depending on the outcome of the review. Until Christiano & Co provide a reliable AI/ML engine, I'd tend to incorporate and maintain these codes into at least the dRICH reconstruction for ePIC, no matter they will be required for pfRICH or not. At that point the standalone pfRICH suite will probably be abandoned for good anyway (but please, only after there is a usable interactive event display in ePIC environment  🙂).

  Cheers,
    Alexander.

________________________________

c-dilks:
> Can you mock up an event with this filled? I'm wondering about the `npe` field in both CerenkovParticleID`and`CerenkovPdgHypothesis`. If you observe, say, 12 pe in an event it goes in `CerenkovParticleID`, and if 11 are predicted for an electron, that goes in the hypothesis, I assume. What about covariance on `npe` for that hypothesis? Presumably you need it for the likelihood.

Perhaps instead I send the [part of the algorithm that fills this collection](https://github.com/eic/EICrecon/blob/d3f2be6a7036ca5fe33a91e97f18043525f6e525/src/algorithms/pid/IrtCherenkovParticleID.cc#L355-L402); this is part of the [big PR](https://github.com/eic/EICrecon/pull/393) to enable IRT PID for the dRICH. I can DM you a ROOT file if you want some numbers; for unrelated reasons my scripts aren't working at the moment, otherwise I would post some plots.

`CherenkovParticleID::npe` counts the number of photoelectrons that were chosen by _any_ mass hypothesis, and is an integer. Upstream of this algorithm we read the digitized hits and assume 1 hit is 1 NPE, which is not always, but usually true; we could use the ADC distribution to try to figure out if a hit corresponds to 2 or more NPE, but the higher priority is that we should switch the digitizer to give us Time over Threshold instead.

As for `CherenkovPdgHypothesis::npe`, this comes from [the IRT algorithm itself](https://github.com/eic/irt/blob/f0e14dd0a2b8fad44fa8a19fe260fe7e537cd539/include/CherenkovPID.h#L43-L45) and it's a float; I don't know the details of how this is determined (I'm only familiar with how to use the IRT algorithm, not its internals), so perhaps @alexander-kiselev could provide more details if needed (if he hasn't already in other comments in this PR).

In general `CherenkovParticleID::npe > CherenkovPdgHypothesis::npe`.

c-dilks:
> So, I'd almost suggest to store the photon energy in GeV, which I know is silly but I know is what geant4 does

Photon energy in GeV is fine with me; the energy deposition of `RawTrackerHit` is also in GeV.

edm4eic.yaml
c-dilks: Personally prefer `PID`, but going with `ParticleID` for consistency.

edm4eic.yaml
```diff
@@ -140,6 +140,13 @@ components:
       - edm4eic::Cov2f    directionError  // Error on the polar and azimuthal angles
       - float             pathlength      // Pathlength from the origin to this point
       - float             pathlengthError // Error on the pathlenght
+
+  ## PID hypothesis from Cherenkov detectors
+  edm4eic::CherenkovPdgHypothesis:
+    Members:
+      - int32_t           pdg             // PDG code
+      - float             npe             // Overall photoelectron count
+      - float             weight          // Likelihood estimate
```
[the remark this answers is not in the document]
c-dilks: @alexander-kiselev or @chchatte92: does IRT's `MassHypothesis::GetWeight` return the likelihood or log(likelihood)? What do we want to store in the data model, likelihood or log(likelihood)?

edm4eic.yaml
c-dilks: Yes, should be unsigned. I don't think we can make it a part of the cell ID, since it's not clear which radiator a pixel hit came from, and in some cases, 2 photons could hit a single pixel, each originating from a different radiator.

edm4eic.yaml
c-dilks: Good idea! I added a 1-N relation to the `MCRecoTrackerHitAssociation`s. This is good for our simulation purposes where we can know the MC truth, but in real data we won't know the truth; in the real-data case, we could still use `MCRecoTrackerHitAssociation` but the `simHits` relation would simply not be filled, so it still works. What do you think?

alexander-kiselev:
Hello colleagues,

  if the idea is that a reconstructed (call it digitized) hit can have indistinguishable contributions from more than one parent particle, this is of course a right approach. Would be typical for calorimetry showers, and was implemented as a rather sophisticated collection of C++ classes in EicRoot, see e.g. here https://github.com/eic/EicRoot/blob/master/eic/calorimetry/EicCalorimeterDigiHit.h#L108). But even pfRICH codes in their infancy have provisions to do the same: https://github.com/eic/irt/blob/pfrich/source/Digitization.cc#L142, namely if two photons hit the same pixel, it is at the end a single hit (perhaps with a better timing), as seen by the ring finder.

  Cheers,
    Alexander.


________________________________

edm4eic.yaml
c-dilks: > This is observed, not estimated, right? If estimated, doesn't it depend on the particle type hypothesis?

Photon theta is estimated from the ray-tracing, independent of hypothesis. 

> Isn't this just avg of the vector members and therefore redundant?

Correct, it is redundant and we can remove this member.

edm4eic.yaml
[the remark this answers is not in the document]
c-dilks: I changed the description to `Weight of this hypothesis, such as likelihood, moment, etc.` to make it more generic. Currently our weights from IRT v1.0.6 are more like moments, but I believe this will change again with the new IRT version (on branch `pfrich` from @alexander-kiselev's pfRICH work, to be eventually released as `v1.1.0+`, not targeting this 'end-of-march' campaign).

c-dilks:
@wdconinc I believe all conversations from above have been resolved, and that this is good enough for a first attempt. From the above discussions, it's clear we will likely rework this anyway, and we should just take this version as a first draft. We need this `datatype` merged in so the IRT usage in EICrecon will run; whether we actually save the corresponding output collections or not is a question that will be answered later (soon), after seeing how much disk space it will take (e.g., we may not want to store all the photons in the full RECO output...)

wdconinc (approved):

PR #19, opened 2022-12-30
fix: allow setting C++ standard to 17|20

### LINKED ISSUE #18: C++20 standard should not be forced
https://github.com/eic/EDM4eic/blob/530ec7ed8aa3b66f5b434a8f5352c1335b1db415/CMakeLists.txt#L9

In reality we only require C++17 and allow C++20. By forcing C++20, we get in trouble with ROOT and nlohmann-json, when including both of these onto the same compilation unit. ROOT redefines its own `span` based on the C++17 standard at ROOT compile time, which conflicts with the `std::span` at EDM4eic compile time.

wdconinc:
Fixes #18 and allows setting C++ standard. Should probably be set to whatever ROOT, podio, and EDM4hep are using, and deviations are at the peril of the developer.

- [x] Bug fix (issue #18)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, C++17 is standard.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,8 +6,13 @@ project(EDM4EIC
   VERSION 1.1.0
   LANGUAGES CXX)
 
-set(CMAKE_CXX_STANDARD 20)
+# C++ standard
+set(CMAKE_CXX_STANDARD 17 CACHE STRING "Set the C++ standard to be used")
+if(NOT CMAKE_CXX_STANDARD MATCHES "17|20")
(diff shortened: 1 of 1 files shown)


wdconinc:
CI failure due to CERN ecsft downtime?

```
2022-12-30T04:49:25.6504387Z curl: (28) Failed to connect to ecsft.cern.ch port 443 after 129734 ms: Connection timed out
```

c-dilks (approved):

PR #23, opened 2023-02-24
feat: add association datatype `MCRecoTrackerHitAssociation`

### LINKED ISSUE #21: add association or relation for `RawPMTHit` to MC
**Is your feature request related to a problem? Please describe.**
For algorithm development it is useful to have the MC photon info, given digitized `RawPMTHit`s. This is a specific need for RICH detectors, where we expect ~1 photon per digitized hit, but sometimes 2 or 3. We thus prefer a one-to-many mapping of hits to photons.

**Describe the solution you'd like**
This is not the kind of mapping we would have available from real data, and generally `MCReco` associations are preferred for these. We need a 1-N mapping:
```yaml
  edm4eic::MCRecoPMTHitAssociation:
    Members:
     - float weight                  // weight of this association
    OneToOneRelations:
     - edm4eic::RawPMTHit rawHit      // reference to the digitized hit
    OneToManyRelations:
     - edm4hep::SimTrackerHit simHits // reference to the simulated hits
```
Note the MC photons are available in the `SimTrackerHit::MCParticle` 1-1 relation.

**Describe alternatives you've considered**
An alternative is a 1-N relation:
```yaml
  edm4eic::RawPMTHit:
     Members:
      - uint64_t          cellID            // The detector specific (geometrical) cell id.
      - uint32_t          integral          // PMT signal integral [ADC]
      - uint32_t          timeStamp         // PMT signal time [TDC]
     OneToManyRelations:
      - edm4hep::SimTrackerHit simHits  // simulated hits
```
This is much more straightforward to handle downstream (and therefore somewhat preferable), but has the downside of introducing a Raw->Truth relation (but is that really so bad?)

**Additional context**
Ideally the RICH PID algorithms do not need MC truth photons; however, during development and testing, they are extremely useful to have.

c-dilks:
Add new one-to-many association `datatype` to link digitized tracker hits (`RawTrackerHit`) to the MC-level `SimTrackerHit`s. Close #21.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
-# Copyright (C) 2022 Sylvester Joosten, Whitney Armstrong, Wouter Deconinck
+# Copyright (C) 2023 Sylvester Joosten, Whitney Armstrong, Wouter Deconinck, Christopher Dilks
 # Some datatypes based on EDM4hep. EDM4hep license applies to those sections.
 ---
 options :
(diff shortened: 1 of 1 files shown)


wdconinc (approved):
This looks good to me. I wonder if you can submit this to EDM4hep as well as a PR, along with a version for SimCalorimeterHit vis à vis RawCalorimeterHit?

c-dilks:
Shall we merge this here now, and upstream things to EDM4hep later?

PR #26, opened 2023-03-12
ci: only run on push main/tags; trigger juggler only on PR

wdconinc:
We were triggering juggler master (incorrect branch) on schedule. That is not necessary.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -2,6 +2,10 @@ name: linux-eic-shell
 
 on:
   push:
+    branches:
+      - main
+    tags:
+      - '"'
   pull_request:
   schedule:
     - cron: '41 4 * * 0'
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -2,6 +2,10 @@ name: linux-eic-shell
 
 on:
   push:
+    branches:
+      - main
+    tags:
+      - '"'
```
wdconinc, line 8:
```suggestion
      - '*'
```

.github/workflows/linux-eic-shell.yml
```diff
@@ -2,6 +2,10 @@ name: linux-eic-shell
 
 on:
   push:
+    branches:
+      - main
+    tags:
+      - '*'
   pull_request:
```
c-dilks, line 9: Every job except for `deploy-docs` (which only runs on `main`) still runs on PRs though.. is that what you want?

wdconinc: Yeah, we still want to test pull requests. Mainly we don't want to trigger juggler builds every day.

c-dilks (approved):

PR #30, opened 2023-04-18
refactor: remove `CherenkovParticleID` member `radiator`

c-dilks:
No longer needed since reconstruction algorithms are now capable of outputting multiple collections.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: feature removal

Yes, but only in a dev branch and I will take care of them

No

### DIFF
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -276,7 +276,6 @@ datatypes:
     Description: "Cherenkov detector PID"
     Author: "A. Kiselev, C. Chatterjee, C. Dilks"
     Members:
-      - uint32_t          radiator          // Radiator number
       - float             npe               // Overall photoelectron count
       - float             refractiveIndex   // Average refractive index at the Cherenkov photons' vertices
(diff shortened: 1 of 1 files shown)


c-dilks:
Relevant EICrecon code has been updated, so this is ready. I don't really need this in `eic-shell`, since not using `setRadiator` or `getRadiator` anywhere should leave it at its default value of `0` and no one will care.

wdconinc (approved):
We should really bunch up our backwards incompatible interface changes like this so we don't have to do two major version changes in a week ... Can we let this linger a bit?

c-dilks:
> We should really bunch up our backwards incompatible interface changes like this so we don't have to do two major version changes in a week ... Can we let this linger a bit?

Sure, but so far I'm the only one using this ;)

We will probably refactor this whole `datatype` anyway... perhaps this PR can eventually just become that refactor, and removal of `radiator` is the first step.

PR #35, opened 2023-05-03
add utils/include/edm4eic/unit_system.h

veprbl:
Defines constants to support EDM4hep's unit system to be adopted as the default for EICrecon.


c-dilks:
Perhaps this would be better in EDM4hep, for everyone's benefit.

veprbl:
Filed downstream PR eic/EICrecon#671 to start using this.

wdconinc:
Only noticed this PR now. This looks good.

wdconinc (approved):

PR #47, opened 2023-09-05
CMakeLists.txt: correct reported version for the next release

veprbl:


wdconinc:
We could have an action do this (here and elsewhere). Some options:
- Update version number upon release,
- Release action where we specify the new release and it commits that change first,
- Or we use `.gitattributes`.

wdconinc (approved):

PR #48, opened 2023-09-05
CMakeLists.txt: use AnyNewerVersion compatibility mode

veprbl:
See https://github.com/eic/EICrecon/pull/903


wdconinc (approved):
This is in some ways making a commitment to being infinitely backwards compatible. All newer versions will work too.

Or it shouldn't be read as such and we reserve the right to break things anyway, and we are maybe only guaranteeing the next newer version in practice, a setting that CMake doesn't offer...

In any case, let's merge it, and move on.

PR #53, opened 2023-09-28
feat: add schema_version

wdconinc:
Podio supports schema evolution. This adds the `schema_version` as an integer composed of major and minor version number. No changes have been added that require a schema evolution specification.

Regrettably, this add yet another location where the version number needs to be kept track off, so we now have:
- CMakeLists.txt project version
- edm4eic.yaml schema_version
- tagged release version

- [ ] Bug fix (issue #__)
- [x] New feature (issue: add schema version)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -5,6 +5,10 @@
 ## Schema versioning
 ## This is required to be an integer, and defined as 100*major+minor.
 ## Patch level changes are required to be schema invariant.
+##
+## If there are schema version changes that can be evolved, see the podio documentation
+## for an example: https://github.com/hegner/podio/blob/master/tests/schema_evolution.yaml
+##
 schema_version: 301
 
 options :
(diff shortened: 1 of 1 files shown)


ruse-traveler:
Looks good! Correct me if I'm wrong, but this is what we decided on at the software meeting last week, right?

wdconinc:
I think last week we did indeed decide on the 100*major+minor integer for schema versioning. We didn't have much discussion about what level of schema evolution we expect; e.g. do we require all field changes to be propagated within minor version changes only? But we can probably look at this as something we decide on a feature-by-feature basis: if your change is not supported by schema evolution, then it becomes inherently API incompatible and requires a new major version, which will probably make it less likely to be adopted in the short term. That would be the pressure to support a clean schema evolution clause to allow use within the current major version and get the feature out sooner.

But that's just me thinking out loud here, and we didn't discuss that part. Maybe we can add it to next Wednesday's discussion which will be already entirely about data model additions :-)

ruse-traveler:
Gotcha! And I like that way of looking at it: since we already require discussion of data model changes, determining the schema evolution level should be part of that discussion. (And agreed on the major versioning!)

ruse-traveler (approved):
LGTM!

ruse-traveler:
(Forgot to hit merge last week 😵)

PR #57, opened 2023-10-18
fix: put all operators in vector_utils.h inside namespace

wdconinc:
This puts all operators inside the namespace edm4eic instead of having part of them in, part of them out. Having some out of the namespace affects other headers included after edm4eic/vector_utils.h, e.g.
```
In file included from /home/wdconinc/git/EICrecon/src/algorithms/tracking/TrackProjector.cc:30:
/opt/._local/kabrpic7bxoos4p6tucb5wr2sce6oeio/include/edm4eic/vector_utils.h:153:34: error: overloaded 'operator/' must have at least one parameter of class or enumeration type
  153 | template <edm4eic::VectorND V> V operator/(const V& v, const double d) {
      |                                  ^
/opt/._local/kabrpic7bxoos4p6tucb5wr2sce6oeio/include/fmt/core.h:1483:29: note: in instantiation of function template specialization 'operator/<int>' requested here
 1483 | enum { max_packed_args = 62 / packed_arg_bits };
      |                             ^
```

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No. I guess you could be relying on these operators being available outside of the edm4eic namespace... but that's off-label use of this product.

### DIFF
--- a/utils/include/edm4eic/vector_utils.h
+++ b/utils/include/edm4eic/vector_utils.h
@@ -116,7 +116,6 @@ template <Vector3D V> double projection(const V& v, const V& v1) {
   return v * v1 / norm;
 }
 
-} // namespace edm4eic
 template <edm4eic::Vector2D V> V operator+(const V& v1, const V& v2) {
   return {edm4eic::vector_x(v1) + edm4eic::vector_x(v2),
           edm4eic::vector_y(v1) + edm4eic::vector_y(v2)};
@@ -153,5 +152,8 @@ template <edm4eic::VectorND V> V operator-(const V& v1, const V& v2) {
 template <edm4eic::VectorND V> V operator/(const V& v, const double d) {
   return (1. / d) * v;
 }
+
+} // namespace edm4eic
+
 #endif
 #endif


veprbl (approved):
edm4hep does this too https://github.com/key4hep/EDM4hep/blob/7977dd05311f6b49639b6d2a97373eed03113070/utils/include/edm4hep/utils/vector_utils.h#L270C2-L270C2

wdconinc:
Yeah, it has another level with utils, but all is in edm4hep at least.

veprbl:
I don't quite understand the error. It seems like this needs just the `explicit` keyword in front of each operator definition instead.

If I understand correctly, users may need to put `using edm4eic::operator/` after this.

veprbl:
I see, the `VectorND` here is a C++ concept. And this only shows up with clang: https://godbolt.org/z/YGa8bW9b8
The `explicit` keyword doesn't apply since this is not a class. Surprisingly, this works
```c++
template <VectorND V, std::floating_point P> V operator/(const V& v, const P d) {
    return v;
}
```

PR #58, opened 2023-10-19
Update track fit EDM

osbornjd:
This introduces changes to the track fit and track state EDM objects to better conform to a single container that can deliver all of the needs of downstream reconstruction algorithms and (ultimately) physics analysis. 

1. Introduces a 6D covariance matrix object that is really the upper triangle of a full 6x6 covariance, for storage purposes.
2. Moves some "track level" quantities out from `edm4eic::Trajectory` and into `edm4eic::Track`
3. Revamps the `edm4eic::TrackParameters` object, which will need to contain the complete covariance to swap between edm4eic and Acts::EDM. Also adds a surface type as bound track parameters make no sense without an associated surface.
4. Adds a full covariance, position, and PID hypothesis to `edm4eic::Track`

The resulting track object contains the set of track parameters at the vertex, some criteria to evaluate the track fit (e.g. those contained in the one to one relation to `edm4eic::Trajectory`), and all track states associated to the fit (also still in `edm4eic::Trajectory`).

The track parameters at the vertex do need to be stored separately (e.g. not as another uniquely defined track state), for a technical reason. They are going to (1) be most frequently used in global coordinates, which the track parameters are not in and (2) are associated to a `PerigeeSurface`, which is transient unlike the surfaces associated with the actual geometry. Without knowledge of the surface, a set of BoundTrackParameters do not have a well defined global definition.

Opening this PR now as it will likely have nontrivial downstream consequences, so comments welcome.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

It introduces breaking changes to EICrecon. Currently there are 3 containers produced by the CKF in EICrecon - the aim of this PR is to reduce this to a single container. This will in turn change other downstream reconstruction algorithms that rely on the CKF output as they will have to access this changed structure.

No


edm4eic.yaml
```diff
@@ -135,6 +135,27 @@ components:
...
+      declaration: "
+        Cov6f() : covariance{} {}\n
+        Cov6f(std::array<float, 21> vcov) : covariance{vcov}{}\n
+	float getCov(unsigned i, unsigned j) const {\n
+	  if(i > j) {\n
+     	    std::swap(i, j);\n
+	  }\n
+	  return covariance[i + 1 + (j + 1) * (j) / 2 - 1];\n
+	}\n
+	void setCov(unsigned i, unsigned j, float val) {\n
```
wdconinc, line 151: This kinda duplicates the same serialization code. Is it possible to have just one that returns a reference, `float& cov(unsigned i, unsigned j)` and then usage is `cov(i,j) = 0.1`?

edm4eic.yaml
```diff
@@ -372,39 +391,40 @@ datatypes:
...
-      - edm4eic::Cov3f    momentumError     // Covariance on theta, phi and qOverP
-      - float             time              // Track time [ns]    
-      - float             timeError         // Error on the time
-      - float             charge            // Particle charge
+      - int32_t              type              // Type of track parameters (-1/seed, 0/head, ...)
+      - uint64_t             surface           // Surface for bound parameters (geometryID)
+      - edm4hep::Vector2f    loc               // 2D location on surface
+      - float                theta             // Track polar angle [rad]
+      - float                phi               // Track azimuthal angle [rad]
+      - float                qOverP            // [e/GeV]
```
wdconinc, line 401: Add pdg hypothesis associated with q, and propagation?

edm4eic.yaml
```diff
@@ -372,39 +391,40 @@ datatypes:
...
+      - int32_t            type                           // Flag that defines the type of track
+      - edm4hep::Vector3f  position                       // Track 3-position at the vertex 
+      - edm4hep::Vector3f  momentum                       // Track 3-momentum at the vertex [GeV]
+      - edm4eic::Cov6f     positionMomentumCovariance     // Covariance matrix in basis [x,y,z,px,py,pz]
+      - float              time                           // Track time at the vertex [ns]
+      - float              timeError                      // Error on the track vertex time
+      - float              charge                         // Particle charge
+      - float              chi2                           // Total chi2
+      - uint32_t           ndf                            // Number of degrees of freedom
+      - uint32_t           pid                            // PDG particle ID hypothesis
```
wdconinc, line 422: Other data types use `pdg` if I recall correctly.

edm4eic.yaml
```diff
@@ -372,39 +391,40 @@ datatypes:
...
     OneToOneRelations:
-      - edm4eic::Trajectory trajectory      // Trajectory of this track
-      - edm4eic::Vertex   vertex            // Track vertex of this track
+      - edm4eic::Trajectory                     trajectory      // Trajectory of this track
     OneToManyRelations:
-      - edm4eic::TrackerHit trackerHits     // Hits that were used for this track
-      - edm4eic::Track    tracks            // Tracks (segments) that have been combined to create this track
+      - edm4eic::TrackerHit trackerHits       // Hits that were used for this track
```
wdconinc, line 426: These should be `Measurement2D` now I think.

edm4eic.yaml
osbornjd: What do you mean by "propagation"?

edm4eic.yaml
osbornjd: Sure, I didn't know if we were ready to move to this

edm4eic.yaml
sly2j: I personally prefer pid, but also like consistency :)

sly2j: PDG code should be `int32_t` (signed integer)

sly2j (changes requested):

edm4eic.yaml
sly2j: Are we using Measurement2D now? Else we should first go to `TrackerHit` to decouple both PRs (but make a note of what it should be, and already add the Measurement2D version commented out).

edm4eic.yaml
```diff
@@ -372,39 +391,40 @@ datatypes:
...
+      - float              time                           // Track time at the vertex [ns]
+      - float              timeError                      // Error on the track vertex time
+      - float              charge                         // Particle charge
+      - float              chi2                           // Total chi2
+      - uint32_t           ndf                            // Number of degrees of freedom
+      - uint32_t           pid                            // PDG particle ID hypothesis
     OneToOneRelations:
-      - edm4eic::Trajectory trajectory      // Trajectory of this track
-      - edm4eic::Vertex   vertex            // Track vertex of this track
+      - edm4eic::Trajectory                     trajectory      // Trajectory of this track
```
sly2j, line 424: Are you still keeping `Trajectory`? I thought we were going straight from TrackParameters to Track now? I guess I'm a little unclear on how you see the data flow?

osbornjd: A track stores a set of track parameters and a trajectory. The track parameters store the track state information at all of the surfaces, the trajectory stores global track quantities like the number of measurements used in the fit, and the Track itself stores the track parameters at the vertex. In principle we could dump the Trajectory entirely and just store this information in the Track itself since they are completely one-to-one, I thought that the distinction was useful though because the Trajectory is really tracking expert information while the Track (that holds the track parameters at the vertex) is really analysis level information.

edm4eic.yaml
osbornjd: I think we are using `TrackerHit` so if we haven't swapped to `Measurement2D` yet then I agree we should decouple these changes.

edm4eic.yaml
wdconinc: No (well, in part; consistency with other Covs is desireable too), I meant returning a reference to the underlying quantity that can then be used in assignment as well, e.g. what you likely know from Acts (e.g. https://github.com/acts-project/acts/blob/1f6dd58d096f89452c1f5d8d5da65b6ddbb7e0bd/Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/ParticleSmearing.cpp#L138), which comes from Eigen at https://eigen.tuxfamily.org/dox/classEigen_1_1DenseCoeffsBase_3_01Derived_00_01WriteAccessors_01_4.html#ae6ba07bad9e3026afe54806fdefe32bb. I think that requires non-const function with return type reference, and can be operator() or a named function.

edm4eic.yaml
wdconinc: It's needed in the Acts::Propagator and Acts::Stepper, which starts from TrackParameters, e.g. through https://github.com/acts-project/acts/blob/1f6dd58d096f89452c1f5d8d5da65b6ddbb7e0bd/Core/include/Acts/EventData/TrackParameters.hpp#L14

edm4eic.yaml
osbornjd: Yeah I think I updated it to what you meant, let me know if that is not true (minus the return by reference, added in my local version)

edm4eic.yaml
wdconinc: We are already using Measurement2D, https://github.com/eic/EDM4eic/pull/54. Maybe this branch needs a rebase?

edm4eic.yaml
osbornjd: Oh you were just referring to the pdg hypothesis, I misunderstood your comment as saying ",and propagation" was referring to something else needed in the `TrackParameters`. You meant that the pid hypothesis is associated to charge and the propagation in Acts - understood now

edm4eic.yaml
osbornjd: Is Measurement2D used anywhere in EICrecon though? I don't see it

wdconinc: It's going to be merged today probably, https://github.com/eic/EICrecon/pull/1076

What matters is what is in this repo for this discussion.

edm4eic.yaml
sly2j: OK, I think I agree. It can become cumbersome if we always have to deal with relations, but if it's expert only it makes sense to separate them (and then have the option of only storing Tracks)

wdconinc:
1. Propagate the Cov6f interface to make other Cov2,3f components identical in API, even if possible regression.
   - this can be addressed in a separate PR 
3. Vector3f for track vertex position is not how we can constrain tracks, and I'd want that in the Vertex
   - Ideally Track is what comes out of CKFTracking, and we cannot add a vertex later through IVF. So what would this be filled with?
5. Cov6f for track position and momentum in cartesian components is unlikely to be calculable.
   - How do we get these cross correlations? Does Acts give those as output from IVF? Does it make sense to have these in cartesian vs spherical coordinates for momentum? Or will we provide a Jacobian to get from one to the other? I like the idea to allow full covariance, but I don't know if this is the coordinate system that makes most sense.

osbornjd:
> 1. Propagate the Cov6f interface to make other Cov2,3f components identical in API, even if possible regression.
>        
>        * this can be addressed in a separate PR

That makes sense.

>     2. Vector3f for track vertex position is not how we can constrain tracks, and I'd want that in the Vertex
>        
>        * Ideally Track is what comes out of CKFTracking, and we cannot add a vertex later through IVF. So what would this be filled with?

This is not a vertex position, this is the track position of closest approach to the vertex (or some, in the case of Acts, `PerigeeSurface` which is right now by default set to (0,0,0) in the CKF). A set of track parameters can't be uniquely defined without a position.

>     3. Cov6f for track position and momentum in cartesian components is unlikely to be calculable.
>        
>        * How do we get these cross correlations? Does Acts give those as output from IVF? Does it make sense to have these in cartesian vs spherical coordinates for momentum? Or will we provide a Jacobian to get from one to the other? I like the idea to allow full covariance, but I don't know if this is the coordinate system that makes most sense.

It is calculable via jacobians that relate the two bases, there are transforms that do this that you can basically copy from the `CovarianceEngine` within Acts. I'll emphasize that this covariance has nothing to do with a vertex, it is on the track position itself WRT a `PerigeeSurface`. My personal opinion is that it does not make sense to store a covariance in local coordinates because it has no physical meaning without an associated surface, and the track parameters at a PerigeeSurface can change depending on where they are propagated (e.g. if they were determined at a primary vs. secondary vertex or something of this nature). Analyzers will, unless I am mistaken, work in global coordinates and the covariance is needed to determine things like the uncertainty on the DCA. We are either going to have to provide the parameters at the perigee surface in global coordinates, or provide them in local coordinates + a surface and additionally provide analyzers with a standard way to transform to global

wdconinc:
Sounds good to me. Up to @sly2j to click the merge button.

osbornjd:
So for my own understanding - this gets merged first and then we make the corresponding changes to EICrecon that will reflect these changes? I'm pretty sure merging this will break the nightly build

wdconinc:
The nightly build in eic-shell doesn't pull in the main branch of just anything, only epic and eicrecon. So merging this doesn't mean it gets into eic-shell. We first need to cut a release here and put it explicitly in eic-shell.

osbornjd:
Thanks for the clarification, I naively assumed all the repos got built together rather than some being only released on versions and some not. That makes sense for repos that don't change as frequently as others

wdconinc:
There's the eternal tension between only putting releases into the environments rolled out to users by default, and putting only rolling main versions into the nightly. I wish we could have a nightly that's just main only but it's not workable and leads to too much breakage. So we've settled on epic and eicrecon main only... And there's pressure even here, in particular when changing acts versions in non-compatible ways wrt eicrecon.

sly2j (approved):

PR #60, opened 2023-11-09
bring measurements back to trajectory

ShujieL:
Move measurements from Track (which is introduced in https://github.com/eic/EDM4eic/pull/58) back to Trajectory, and add back outliers.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x ] Other: partially revert a change

- [x ] Changes have been communicated to collaborators.  -- This change is communicated to Joe Osborn. 

CKFtracking.cc needs to be updated to take the new data structure.


wdconinc:
We discussed this in Joe's talk. We don't want these hits in trajectory.

osbornjd:
Shujie discussed with me offline and wanted this because (correct me if I'm wrong)
1. It may be confusing to users who use this 
2. The differentiation between outlier measurements and track measurements is needed, even though outliers are a subset of the total measurements that belong to a track. Acts determines what a track state looks like with a flag (basically an enum) that labels it as measurement, outlier, hole, whatever. In principle doing something like this would consolidate the information.

I still personally think we should have the measurements in the `Track` object because analyzers may want to know how many measurements are on a track for quality cuts (e.g. if an analysis requires the best momentum resolution possible, require more measurements on the track). The point of differentiating between `Track` and `Trajectory` was to separate expert from user information, so that is why I put the measurement list in the `Track` object.

wdconinc:
> Shujie discussed with me offline and wanted this because

Whatever it is, let's try to get this right in fewer steps, ok? I.e. this better be the final one.

osbornjd (dismissed):
I don't have an issue with this, my opinion is that the measurements belong in the `Track` object as analyzers may want to know how many exist on a track (and how many from various subsystems exist), and if we are trying to keep analyzer info from expert info separate then this should be moved. But I'm not going to argue for it if others think it is better in this proposed way

ShujieL:
I noticed this PR is still open.
The "Track" data structure is not used anywhere in EICrecon. All ongoing analysis access measurements etc from "Trajectory" with the previous tagged version. To be consistent, can we keep this trajectory structure unchanged (i.e. bring the references to hits back as I suggested in this PR). We can always change this when future algorithm development with "Track" is ready. 
@wdconinc @osbornjd

wdconinc:
> The "Track" data structure is not used anywhere in EICrecon.

I think we(you) should address that issue instead of reverting data model changes.

edm4eic.yaml
```diff
@@ -419,7 +421,6 @@ datatypes:
     OneToOneRelations:
       - edm4eic::Trajectory                     trajectory      // Trajectory of this track
     OneToManyRelations:
-      - edm4eic::Measurement2D measurements      // Measurements that were used for this track
       - edm4eic::Track      tracks            // Tracks (segments) that have been combined to create this track
```
wdconinc, line 424: This is the problematic part here. How can we get people to start using the new Track structures when they get implemented if you will have removed the functionality.

Suggestion:
- put `edm4eic::Track::measurements` back,
- rename `edm4eic::Trajectory::{measurements,outliers}` to append `_deprecated` and indicate in the comment what should be used.

wdconinc (changes requested):

edm4eic.yaml
ShujieL: I will do what you suggested. 

But I think we need a bit more discussion on this. In the previous trajectory structure (which I try to revive), we store good measurements (which used in track fit) and outliers (not included in track fit due to chi2 cut etc but is the best candidate on that layer).  In the current analysis we can learn from the outlier position and chi2 on how to improve tracking.  
In this new track structure, we only have one member "measurements". If that means we won't store outliers, then we miss this piece of useful info. If we want to store both track measurements and outliers into it, then we need to add an array of state flag to distinguish them.

edm4eic.yaml
```diff
@@ -385,7 +385,9 @@ datatypes:
       - float             measurementChi2   // Chi2 for each of the measurements
       - float             outlierChi2       // Chi2 for each of the outliers
     OneToManyRelations:
-      - edm4eic::TrackParameters trackParameters // Associated track parameters, if any
+      - edm4eic::TrackParameters trackParameters            // Associated track parameters, if any
+      - edm4eic::Measurement2D measurements_depracated      // Measurements that were used for this track. Will move this to the edm4eic::Track
+      - edm4eic::Measurement2D outliers_depracated          // Measurements that were not used for this track. Will move this to the edm4eic::Track
```
wdconinc, line 390:
```suggestion
      - edm4eic::Measurement2D measurements_deprecated      // Measurements that were used for this track. Will move this to the edm4eic::Track
      - edm4eic::Measurement2D outliers_deprecated          // Measurements that were not used for this track. Will move this to the edm4eic::Track
```

wdconinc (approved):

PR #61, opened 2023-12-12
feat: update vertex edm to 4d quantities

osbornjd:
This PR modifies the vertex EDM to contain 4D information, which is going to be needed in our streaming readout environment to separate pile up from physics bunch crossings. 

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Discussed in S&C meeting on the 18th of October [here](https://indico.bnl.gov/event/20894/).
It changes the vertex object for which the vertexing algorithm in EICRecon will need to be updated. This should be a simple fix since nothing else depends on the vertexing algorithm at the moment as it is still under development.

### DIFF
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -438,7 +438,7 @@ datatypes:
       - int32_t             type          // Type flag, to identify what type of vertex it is (e.g. primary, secondary, generated, etc.)
       - float               chi2          // Chi-squared of the vertex fit
       - int                 ndf           // NDF of the vertex fit
-      - edm4hep::Vector4f   position      // [mm] position + time t0 of the vertex. Time is 4th component in vector
+      - edm4hep::Vector4f   position      // position [mm] + time t0 [ns] of the vertex. Time is 4th component in vector
       ## this is named "covMatrix" in EDM4hep, renamed for consistency with the rest of edm4eic
       - edm4eic::Cov4f      positionError // Covariance matrix of the position+time. Time is 4th component, similarly to 4vector 
     OneToManyRelations:
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -436,7 +436,7 @@ datatypes:
     Members:
       - int32_t             type       // Type flag, to identify what type of vertex it is (e.g. primary, secondary, generated, etc.)
       - float               chi2          // Chi-squared of the vertex fit
-      - int               ndf   // NDF of the vertex fit
+      - int                 ndf   // NDF of the vertex fit
       - edm4hep::Vector4f   position      // [mm] position + time t0 of the vertex.
       ## this is named "covMatrix" in EDM4hep, renamed for consistency with the rest of edm4eic
       - edm4eic::Cov4f      positionError // Covariance matrix of the position+time 
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -437,7 +437,7 @@ datatypes:
     Members:
       - int32_t             type       // Type flag, to identify what type of vertex it is (e.g. primary, secondary, generated, etc.)
       - float               chi2          // Chi-squared of the vertex fit
-      - int                 ndf   // NDF of the vertex fit
+      - int                 ndf           // NDF of the vertex fit
       - edm4hep::Vector4f   position      // [mm] position + time t0 of the vertex.
       ## this is named "covMatrix" in EDM4hep, renamed for consistency with the rest of edm4eic
       - edm4eic::Cov4f      positionError // Covariance matrix of the position+time 
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -149,7 +149,6 @@ components:
 	  return covariance[i + 1 + (j + 1) * (j) / 2 - 1];\n
 	}\n
       "
- 
   ## A point along a track
   edm4eic::TrackPoint:
     Members:
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -149,7 +149,20 @@ components:
 	  return covariance[i + 1 + (j + 1) * (j) / 2 - 1];\n
 	}\n
       "
-
+  edm4eic::Vector4f:
+    Members:
+      - float x
+      - float y
+      - float z
+      - float t
+    ExtraCode:
+      declaration: "
+      constexpr Vector4f() : x(0),y(0),z(0),t(0) {}\n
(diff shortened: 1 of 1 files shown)


osbornjd:
Also opened an issue [here](https://github.com/key4hep/EDM4hep/issues/245) for including 4d vectors in edm4hep

osbornjd:
Opened a PR upstream for which this can eventually take advantage of
https://github.com/key4hep/EDM4hep/pull/248

osbornjd:
@wdconinc this is ready to be merged now once we update to the new edm4hep version with https://github.com/key4hep/EDM4hep/pull/248 merged. The CI here will naturally fail until we update edm4hep versions.

wdconinc:
- [ ] updating edm4hep in https://github.com/eic/containers/pull/8.

edm4eic.yaml
```diff
@@ -149,7 +149,7 @@ components:
 	  return covariance[i + 1 + (j + 1) * (j) / 2 - 1];\n
 	}\n
       "
-
+
```
wdconinc, line 152:
```suggestion

```

edm4eic.yaml
```diff
@@ -433,22 +433,16 @@ datatypes:
...
-    Author: "W. Armstrong, S. Joosten, based off EDM4hep"
+    Author: "J. Osborn"
     Members:
-      - int32_t             primary       // Boolean flag, if vertex is the primary vertex of the event
+      - int32_t             type       // Type flag, to identify what type of vertex it is (e.g. primary, secondary, generated, etc.)
       - float               chi2          // Chi-squared of the vertex fit
-      - float               probability   // Probability of the vertex fit
-      - edm4hep::Vector3f   position      // [mm] position of the vertex.
+      - int               ndf   // NDF of the vertex fit
+      - edm4hep::Vector4f   position      // [mm] position + time t0 of the vertex.
```
wdconinc, line 441:
```suggestion
      - edm4hep::Vector4f   position      // position [mm] + time [ns] of the vertex
```

edm4eic.yaml
```diff
@@ -433,22 +433,16 @@ datatypes:
...
     Description: "EIC vertex"
-    Author: "W. Armstrong, S. Joosten, based off EDM4hep"
+    Author: "J. Osborn"
     Members:
-      - int32_t             primary       // Boolean flag, if vertex is the primary vertex of the event
+      - int32_t             type       // Type flag, to identify what type of vertex it is (e.g. primary, secondary, generated, etc.)
       - float               chi2          // Chi-squared of the vertex fit
-      - float               probability   // Probability of the vertex fit
-      - edm4hep::Vector3f   position      // [mm] position of the vertex.
+      - int               ndf   // NDF of the vertex fit
```
wdconinc, line 440:
```suggestion
      - int                 ndf   // NDF of the vertex fit
```

edm4eic.yaml
```diff
@@ -433,22 +433,16 @@ datatypes:
 
   edm4eic::Vertex:
     Description: "EIC vertex"
-    Author: "W. Armstrong, S. Joosten, based off EDM4hep"
+    Author: "J. Osborn"
     Members:
-      - int32_t             primary       // Boolean flag, if vertex is the primary vertex of the event
+      - int32_t             type       // Type flag, to identify what type of vertex it is (e.g. primary, secondary, generated, etc.)
```
wdconinc, line 438: Can we define the types through getters in `ExtraCode`? See e.g. https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L188-L220 and the respective `MutableExtraCode` above it.

osbornjd: This probably makes sense but if it is okay with you I'd rather defer this to another PR. I don't think we have discussed all of the different types we need to be able to support. This requires input from S&C, background, tracking, and probably others (HF working group...). I can think of a starting list but it is not immediately clear to me that this will be exhaustive.

wdconinc: Ok, sounds good to keep out of scope here. Can you create an issue about that, post it here, and resolve the conversation?

osbornjd: Done https://github.com/eic/EDM4eic/issues/62

edm4eic.yaml
```diff
@@ -433,22 +433,16 @@ datatypes:
...
       - float               chi2          // Chi-squared of the vertex fit
-      - float               probability   // Probability of the vertex fit
-      - edm4hep::Vector3f   position      // [mm] position of the vertex.
+      - int                 ndf   // NDF of the vertex fit
```
wdconinc, line 440:
```suggestion
      - int                 ndf           // NDF of the vertex fit
```

edm4eic.yaml
wdconinc: Do we want to store the time t0 in units of `mm`? That would imply `ct` and is used in some data models. We should be specific.

osbornjd: That was never discussed in the working group discussions, so the short answer is I don't know. Is there any precedent elsewhere in our EDM to use units of `mm` for time? I would personally default to time since it is more fundamental, i.e. you can always convert to a distance assuming `ct` but if for whatever reason `c` was not the correct velocity you couldn't convert back to `t`

osbornjd: I default to 4th only because that is what Acts does, which isn't necessarily a good reason but would offer consistency. I'll add a comment

edm4eic.yaml
```diff
@@ -435,12 +435,12 @@ datatypes:
     Description: "EIC vertex"
     Author: "J. Osborn"
     Members:
-      - int32_t             type       // Type flag, to identify what type of vertex it is (e.g. primary, secondary, generated, etc.)
+      - int32_t             type          // Type flag, to identify what type of vertex it is (e.g. primary, secondary, generated, etc.)
       - float               chi2          // Chi-squared of the vertex fit
       - int                 ndf           // NDF of the vertex fit
-      - edm4hep::Vector4f   position      // [mm] position + time t0 of the vertex.
+      - edm4hep::Vector4f   position      // [mm] position + time t0 of the vertex. Time is 4th component in vector
```
wdconinc, line 441:
```suggestion
      - edm4hep::Vector4f   position      // position [mm] + time t0 [ns] of the vertex. Time is 4th component in vector
```

wdconinc (approved):

PR #64, opened 2024-01-19
fix: only compile edm4eic_merge for podio < 0.17.4

wdconinc:
With podio 0.17.4, we can't compile emd4eic_merge anymore since it is reliant on podio/EventStore.h. We have been ignoring the deprecation warnings about this for half a year now...

This PR makes the compilation of edm4eic_merge conditional on being before that version. It renames the source and executable to event_merge.cpp and edm4eic_event_merge to make clear this isn't working for frames. Code is not removed since an adventurous soul may port this to frames so we can do background mixing.

- [x] Bug fix (issue: edm4eic doens't compile with podio 0.17.4)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, edm4eic_merge is renamed, and will not be installed when newer versions of podio are used.


veprbl:
Can we just remove this. I don't see how eventstore logic will help revive this.

wdconinc:
I was thinking about that but we don't have any other podio merging code elsewhere (might be something that should go upstream). At this point we've mostly been focusing background efforts on hepmc3 merging, but the goal was to transition to merging after simulation at some point. This could be something that someone can start from.

So, how about we upstream it as a general podio tool and then remove it here?

veprbl:
I'm not sure why do we need to work on a tool that no one uses. Likely, for merging, hadd should be used. Its output may be slightly broken in metadata, but for that, podio can be fixed.

wdconinc:
`hadd` merges one event with one event across input files. It doesn't do one to N, and it doesn't allow N to vary across files, and it doesn't allow N to be a Poisson distributed random variable. Even as a starting point, I wouldn't recommend `hadd` because you would find yourself be limited very quickly.

veprbl (approved):

PR #70, opened 2024-01-31
Add association between tracker hit and mcparticles

ShujieL:
Add association between tracker hit and mcparticles to allow easy trace back from hits to mc particles in analysis.  

- [ ] Bug fix (issue #__)
- [ x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x ] Changes have been communicated to collaborators

No breaking changes.  This is a standalone data structure. Corresponding factory will be developed and inserted to the downstream of tracking algorithm to fill the data structure. It should not interfere with any existing structure.

No.

### DIFF
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -544,3 +544,15 @@ datatypes:
      - edm4eic::RawTrackerHit rawHit        // reference to the digitized hit
     OneToManyRelations:
      - edm4hep::SimTrackerHit simHits       // reference to the simulated hits
+
+
+  edm4eic::MCParticleRecoHitAssociation:
+    Description: "Association between a TrackerHit and MCParticles. One tracker hit can be linked to multiple simulation hits hence multiple particles"
+    Author : "S. Li"
+    VectorMembers:
+      - uint32_t            simIDs          // Index of corresponding MCParticles (position in MCParticles array)
+      - float               weights         // weight of this association
+    OneToOneRelations:
+      - edm4eic::TrackerHit trackHit        // reference to the hit
+    OneToManyRelations:
+      - edm4hep::MCParticle mcParticles     // reference to the Monte-Carlo particle
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -549,9 +549,8 @@ datatypes:
   edm4eic::MCParticleRecoHitAssociation:
     Description: "Association between a TrackerHit and MCParticles. One tracker hit can be linked to multiple simulation hits hence multiple particles"
     Author : "S. Li"
-    VectorMembers:
-      - float               weights         // weight of this association
+    Members:
+      - float               weight         // weight of this association
     OneToOneRelations:
       - edm4eic::TrackerHit trackHit        // reference to the hit
-    OneToManyRelations:
-      - edm4hep::MCParticle mcParticles     // reference to the Monte-Carlo particle
+      - edm4hep::MCParticle mcParticle     // reference to the Monte-Carlo particle
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -550,7 +550,6 @@ datatypes:
     Description: "Association between a TrackerHit and MCParticles. One tracker hit can be linked to multiple simulation hits hence multiple particles"
     Author : "S. Li"
     VectorMembers:
-      - uint32_t            simIDs          // Index of corresponding MCParticles (position in MCParticles array)
       - float               weights         // weight of this association
     OneToOneRelations:
       - edm4eic::TrackerHit trackHit        // reference to the hit


wdconinc (commented):
I'd rather we connect TrackerHit to RawTrackerHit.

edm4eic.yaml
```diff
@@ -544,3 +544,15 @@ datatypes:
...
     OneToManyRelations:
      - edm4hep::SimTrackerHit simHits       // reference to the simulated hits
+
+
+  edm4eic::MCParticleRecoHitAssociation:
+    Description: "Association between a TrackerHit and MCParticles. One tracker hit can be linked to multiple simulation hits hence multiple particles"
+    Author : "S. Li"
+    VectorMembers:
+      - uint32_t            simIDs          // Index of corresponding MCParticles (position in MCParticles array)
+      - float               weights         // weight of this association
```
veprbl, line 554: This is getting strange. The original motivation for keeping simID was to keep poor users from having to access branches with strange names. In this case, using vector members will create `CollectionName.simIDs_begin`, `CollectionName.simIDs_end` and `_CollectionName_simIDs`, so a user would need to do two dereferences instead of one with just `mcParticles_begin` and `mcParticles_end`.

```suggestion
    VectorMembers:
      - float               weights         // weight of this association
```

edm4eic.yaml
```diff
@@ -544,3 +544,15 @@ datatypes:
      - edm4eic::RawTrackerHit rawHit        // reference to the digitized hit
     OneToManyRelations:
      - edm4hep::SimTrackerHit simHits       // reference to the simulated hits
+
+
+  edm4eic::MCParticleRecoHitAssociation:
```
veprbl, line 549: If the association is known at the time when `edm4eic::TrackerHit` is originally created, just add the `mcParticles` field to it, don't make another collection.

ShujieL: This uses raw hit, we want tracker hits. I don't think it's a good ideas to add the mcparticle here and then create another layer of link to go from tracker hit to raw hits. Let's just have a new association to provide the direct tracker hits to mcparticles link which is exactly what will be used in analysis.

veprbl: I don't understand your explanation. Line 555 reads TrackerHit. Are you responding to @wdconinc's comment?

ShujieL: Ahh sorry I thought you suggested to use the existing association structure on line 546 and add mcparticles there. If you mean adding mcparticles directly to trackerhit instead of creating a separate association, then we are having our reconstructing algorithm directly linked to mc information, which is not recommended according to Wouter

ShujieL:
> I'd rather we connect TrackerHit to RawTrackerHit.

I thought about this option. But we don't have the raw hit to sim hit association filled.  So it's almost the same amount of work to add it in EICrecon. 
And if doing so, it takes the user three steps: tracker hit -> raw hit -> sim hit -> mcparticles to know what particle creates the tracker hit. I think it's too much a burden for analysis, so I prefer to have it as an association to be stored in EICrecon output. It is harmless. I will provide a factory to fill in the structure so user can easily access it.

edm4eic.yaml
veprbl: That's a good point. Keeping MC associations as separate collections allows to be very explicit about whether those valuables are available. However, optional factory inputs/outputs are not implemented and there is no clear understanding of how that should work even, so this policy may not be very sound in the long run.

I will mark this as resolved.

wdconinc:
> But we don't have the raw hit to sim hit association filled. So it's almost the same amount of work to add it in EICrecon.
And if doing so, it takes the user three steps: tracker hit -> raw hit -> sim hit -> mcparticles to know what particle creates the tracker hit. I think it's too much a burden for analysis[...]

We should use associations to connect quantities that we can determine from data to truth MC information which is only sometimes available and ONLY when there is a clear one-to-one connection to truth information. There seems to be this desire to create associations to MCParticles for every reconstructed entity there is. Most likely that is not possible and will require that someone makes a choice as to what is the truth which others will disagree with. If an entity is derived from multiple other entities that are uniquely associated with a specific truth entity, then we should not replicate this dependency.

edm4eic.yaml
```diff
@@ -544,3 +544,14 @@ datatypes:
...
+    Description: "Association between a TrackerHit and MCParticles. One tracker hit can be linked to multiple simulation hits hence multiple particles"
+    Author : "S. Li"
+    VectorMembers:
+      - float               weights         // weight of this association
+    OneToOneRelations:
+      - edm4eic::TrackerHit trackHit        // reference to the hit
+    OneToManyRelations:
+      - edm4hep::MCParticle mcParticles     // reference to the Monte-Carlo particle
```
veprbl, line 557: I thought about this. I think the original reason for the `weight` field is that there can be multiple associations. There may be several entries in the association collection, and weight can encode the fractional contribution to the hit/cluster.
```suggestion
    Members:
      - float               weight         // weight of this association
    OneToOneRelations:
      - edm4eic::TrackerHit trackHit        // reference to the hit
      - edm4hep::MCParticle mcParticle     // reference to the Monte-Carlo particle
```
cc @sly2j @whit2333 for confirmation

ShujieL:
> > But we don't have the raw hit to sim hit association filled. So it's almost the same amount of work to add it in EICrecon.
> > And if doing so, it takes the user three steps: tracker hit -> raw hit -> sim hit -> mcparticles to know what particle creates the tracker hit. I think it's too much a burden for analysis[...]
> 
> We should use associations to connect quantities that we can determine from data to truth MC information which is only sometimes available and ONLY when there is a clear one-to-one connection to truth information. There seems to be this desire to create associations to MCParticles for every reconstructed entity there is. Most likely that is not possible and will require that someone makes a choice as to what is the truth which others will disagree with. If an entity is derived from multiple other entities that are uniquely associated with a specific truth entity, then we should not replicate this dependency.

As Dmitry @veprbl suggested, we can make the tracker hit and mcparticle relation one to one, and using the field "weight" to handle the multiple associations. In this way, the relation between truth info and recon quantity is clear and no built-in "choices". 

This is the only association data structure the tracking study requested so far b/c it's necessary for purity calculation. Any alternative solution means extensive local algorithm development out side of EICrecon that is difficult to track and maintain.

veprbl:
> > But we don't have the raw hit to sim hit association filled. So it's almost the same amount of work to add it in EICrecon.
> > And if doing so, it takes the user three steps: tracker hit -> raw hit -> sim hit -> mcparticles to know what particle creates the tracker hit. I think it's too much a burden for analysis[...]
> 
> We should use associations to connect quantities that we can determine from data to truth MC information which is only sometimes available and ONLY when there is a clear one-to-one connection to truth information. There seems to be this desire to create associations to MCParticles for every reconstructed entity there is. Most likely that is not possible and will require that someone makes a choice as to what is the truth which others will disagree with. If an entity is derived from multiple other entities that are uniquely associated with a specific truth entity, then we should not replicate this dependency.

What @ShujieL described last week, to me, sounded as very similar to CalorimeterClusters and MCParticles associations. If we were to follow the principle quoted above, we'd need to remove associations to CalorimeterClusters and introduce one with CalorimeterHit, which is probably not something we want to do. So that's a counterexample.

ShujieL:
> > > But we don't have the raw hit to sim hit association filled. So it's almost the same amount of work to add it in EICrecon.
> > > And if doing so, it takes the user three steps: tracker hit -> raw hit -> sim hit -> mcparticles to know what particle creates the tracker hit. I think it's too much a burden for analysis[...]
> > 
> > 
> > We should use associations to connect quantities that we can determine from data to truth MC information which is only sometimes available and ONLY when there is a clear one-to-one connection to truth information. There seems to be this desire to create associations to MCParticles for every reconstructed entity there is. Most likely that is not possible and will require that someone makes a choice as to what is the truth which others will disagree with. If an entity is derived from multiple other entities that are uniquely associated with a specific truth entity, then we should not replicate this dependency.
> 
> What @ShujieL described last week, to me, sounded as very similar to CalorimeterClusters and MCParticles associations. If we were to follow the principle quoted above, we'd need to remove associations to CalorimeterClusters and introduce one with CalorimeterHit, which is probably not something we want to do. So that's a counterexample.

Thanks @veprbl. In this case, can you or @wdconinc approve it?

wdconinc:
This was discussed last week at the software and computing meeting, https://indico.bnl.gov/event/22292/. There was no agreement and plenty of discussion about whether this is the right way forward or not. There was no conclusion on this. We should not merge this until we reach a better conclusion. You should ask @mdiefent or @sly2j for what and when the next step is.

ShujieL:
To close the loop, the recommendation from the software group was to NOT having this direct association.

Instead, the raw hits and sim hits are added to the EICrecon output. See https://github.com/eic/EICrecon/pull/1301

So with the above PR and the existing sim hits to raw hits association data structure, one can trace back from tracker hits -> raw hits -> sim hits -> mc particle offline in the analysis script. The raw to sim hits association needs to be filled in EICrecon still.

PR #77, opened 2024-04-19
edm4eic.yaml: fix MCRecoTrackerHitAssociation to relate to a single sim hit at a time

veprbl:
I believe, having sim hits as a OneToMany relation was a mistake. The idea behind having a weight is to specify level of contribution of a specific sim hit. If multiple sim hits are associated, user should expect multiple association records in the collection. We are only now introducing the actual use of this, so now is the good time to do a fix https://github.com/eic/EICrecon/pull/1392#discussion_r1572964045

cc @wdconinc @c-dilks


c-dilks:
For the dRICH, `simHits.size > 1` when two optical photons hit the same SiPM pixel, and `weight` is not used.

This PR requires:
- check for anything that assumes that one `MCRecoTrackerHitAssociation` corresponds to one _unique_ raw hit, namely the EICrecon algorithm `IrtCherenkovParticleID`; update if needed
- update `PhotoMultiplierHitDigi`, which fills the collection
- updates may also be needed downstream of EICrecon, namely `reconstruction_benchmarks/benchmarks/rich`

All of these are feasible; please followup with @chchatte92 and @bleaktwig

veprbl:
Thanks for overview, @c-dilks, I've submitted a PR to update EICrecon and https://eicweb.phy.anl.gov/EIC/benchmarks/reconstruction_benchmarks/-/merge_requests/327

c-dilks:
Awesome, thanks!

wdconinc (approved):
Yes, this looks like the right direction to me. Also consistent with EDM4hep where all associations are one-to-one, https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L490-L560.

veprbl:
If there are no objections to this, let's merge.

PR #79, opened 2024-05-07
edm4eic.yaml: bump schema_version to 600

veprbl:


veprbl:
We should release 6.0.1 with this

wdconinc (approved):
We definitely need to make this more automated, like increment it as part of a release action.

PR #82, opened 2024-05-30
feat: new component edm4eic::Surface

wdconinc:
### Briefly, what does this PR introduce?
This PR adds a new component to the datamodel, `edm4eic::Surface`, identical to the ActsPodioEdm component. It allows for persistence of arbitrary surfaces, including surfaces that may change from event to event (perigee surfaces) or otherwise are not available in the geometry (effective surfaces with a depth in the calorimeter determined by the reconstructed energy).

The most immediate use is to allow reuse of this component in an Acts::PodioTrackContainer with edm4eic::TrackCollections as the underlying storage backend.

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #83, opened 2024-05-30
Revert "feat: new component edm4eic::Surface (#82)"

wdconinc:
This reverts commit 3b4975588356d81321b396490fe67fe31e438ac1. See https://github.com/eic/EDM4eic/pull/82#issuecomment-2140915867.


veprbl:
Thanks for your diligence! I don't think this is necessary right now.

PR #84, opened 2024-06-20
CMakeLists.txt: extend podio major versions range

veprbl:
This is to support PODIO v01-00

Fixes:
```
CMake Error at CMakeLists.txt:31 (find_package):
  Could not find a configuration file for package "podio" that is compatible
  with requested version "0.15".

  The following configuration files were considered but not accepted:

    /lib/cmake/podio/podioConfig.cmake, version: 1.0.0
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,7 @@ endif(APPLE)
 
 include(GNUInstallDirs)
 
-find_package(podio 0.15 REQUIRED)
+find_package(podio 0.15...999 REQUIRED)
 include_directories(${podio_INCLUDE_DIR})
 
 find_package(EDM4HEP 0.10.3 REQUIRED)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -28,7 +28,10 @@ endif(APPLE)
 
 include(GNUInstallDirs)
 
-find_package(podio 0.15 REQUIRED)
+find_package(podio 0.15)
```
wdconinc, line 31: Could we use this newer syntax?
```suggestion
find_package(podio 0.15...1.0)
```
See https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature, as of cmake 3.19.

I think upgrading to 3.19 as a requirement is reasonable, since we already need that on mac anyway:
https://github.com/root-project/root/blob/c33e6a4d897ac1dfe7a80c9985c485d90d09be07/CMakeLists.txt#L8

veprbl: See the first commit of the PR (it doesn't work).

wdconinc: I guess that makes sense, in hindsight.

wdconinc (approved):

PR #90, opened 2024-08-02
Update major version to 7

rahmans1:
Update major version to 7

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __


wdconinc (approved):
Looks good.

Out of scope:
- Can we pull the schema_version in the CMakeLists.txt and turn it into the project version? That would be one fewer place to keep track of it (we would not be able to expose the patch number this way).
- Can we write a reliable CI job that checks whether the pre-commit version can still read a post-commit file? Does podio have a "write one default copy of everything" that we can use?

veprbl (approved):

PR #93, opened 2024-08-27
feat: add q four-momentum vector

wdconinc:
This PR adds the four-momentum vector.

### DIFF
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -9,7 +9,7 @@
 ## If there are schema version changes that can be evolved, see the podio documentation
 ## for an example: https://github.com/hegner/podio/blob/master/tests/schema_evolution.yaml
 ##
-schema_version: 700
+schema_version: 701
 
 options :
   # should getters / setters be prefixed with get / set?
@@ -480,6 +480,7 @@ datatypes:
       - float             W                 // Invariant mass of final state [GeV]
       - float             y                 // Inelasticity (P.q/P.k)
       - float             nu                // Energy transfer P.q/M [GeV]
+      - edm4hep::Vector4f q                 // Four-momentum transfer vector
     OneToOneRelations:
       - edm4eic::ReconstructedParticle scat // Associated scattered electron (if identified)
       ## @TODO: Spin state?


wdconinc:
Discussion in zoom chat:
- Can you provide the definitions of q in the case of the other methods, like DA, Sigma, eSigma? As a comment on that PR would be great.
- For SIDIS measurements, you want to define the q-vector w.r.t to the virtual photon direction.
- Yes, but how do you define this in methods that don't calculate this? E.g. https://github.com/eic/EICrecon/blob/main/src/algorithms/reco/InclusiveKinematicsESigma.cc#L85-L94. We want some invariants (Lorentz even) for the data model, such as q.dot(q) = Q2, regardless of which methods are used.
- It's clear for electron method where we explicitly calculate q, https://github.com/eic/EICrecon/blob/main/src/algorithms/reco/InclusiveKinematicsElectron.cc#L127.
- But the other methods... less clear to me.

rseidl-rcf:
> Discussion in zoom chat:
> 
>     * Can you provide the definitions of q in the case of the other methods, like DA, Sigma, eSigma? As a comment on that PR would be great.
> 
>     * For SIDIS measurements, you want to define the q-vector w.r.t to the virtual photon direction.
> 
>     * Yes, but how do you define this in methods that don't calculate this? E.g. https://github.com/eic/EICrecon/blob/main/src/algorithms/reco/InclusiveKinematicsESigma.cc#L85-L94. We want some invariants (Lorentz even) for the data model, such as q.dot(q) = Q2, regardless of which methods are used.
> 
>     * It's clear for electron method where we explicitly calculate q, https://github.com/eic/EICrecon/blob/main/src/algorithms/reco/InclusiveKinematicsElectron.cc#L127.
> 
>     * But the other methods... less clear to me.

Yes, that was a little too naive by me. I need to look up how to do that for DA, and xSIGMA methods. For JB your qvector is just the sum of all hadronic final state 4-momenta minus the incoming hadron 4-momentum.

For clarification, we need to calculate angles and pt relative to the virtual photon direction (ie q vector) and z requires dot products with the q vector.

simonge:
> maybe the addition should be Vector4f for the four-momentum vector, and a Cov4f for the covariance. Thoughts?

Yes, that sounds more reasonable to keep the overall data smaller and simpler.

edm4eic.yaml
```diff
@@ -538,6 +538,7 @@ datatypes:
       - float             W                 // Invariant mass of final state [GeV]
       - float             y                 // Inelasticity (P.q/P.k)
       - float             nu                // Energy transfer P.q/M [GeV]
+      - edm4hep::Vector4f q                 // Four-momentum transfer vector (E, px, py, pz) [GeV]
     OneToOneRelations:
       - edm4eic::ReconstructedParticle scat // Associated scattered electron (if identified)
```
simonge:
```suggestion
      - edm4hep::Vector4f q                 // Four-momentum transfer vector (E, px, py, pz) [GeV]
      - edm4eic::Cov4f    qError            // Covariance matrix of the Four-momentum transfer vector (10 parameters).
    OneToOneRelations:
      - edm4eic::ReconstructedParticle scat          // Associated scattered electron (if identified)
      - edm4eic::HadronicFinalState    hadronicState // Associated hadonic final state (if used in calculation)
```

simonge (dismissed):
LGTM - might be worth another opinion too.

simonge:
Probably not worth including qError at the moment then while the propagation cannot be done properly. A lookup table could be produced just comparing the true q values to the ones calculated by the reconstruction method.

What could also be useful is the momentum and uncertainty on the beam vectors used in the calculation.

PR #100, opened 2025-01-20
feat: deprecate simID and recID accessors

wdconinc:
This deprecates the `simID` and `recID` fields of the associations, instead of just simply removing them entirely in #51.

See backwards compatibility note below.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: deprecate duplicated information)
- [ ] Documentation update
- [ ] Other: __

This will still break ROOT scripts that directly access the data storage layer (leafs in ROOT tree) in the ROOT files. In podio context that's a private interface, but many in EIC have been treating it as public.

Yes, `simID` and `recID` will not be stored.

### DIFF
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -513,45 +513,81 @@ datatypes:
     Description: "Used to keep track of the correspondence between MC and reconstructed particles"
     Author: "S. Joosten"
     Members:
-      - uint32_t          simID             // Index of corresponding MCParticle (position in MCParticles array)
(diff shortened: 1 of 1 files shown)


ShujieL:
seems that recID and simID are still actively used here https://github.com/eic/EICrecon/blob/32070b472cad3bdc10f957e1ff43273c56118a80/src/algorithms/calorimetry/EnergyPositionClusterMerger.h#L154. Would that be a problem?

wdconinc:
Deprecation is in part a strategy to manifest residual usage.

ShujieL (approved):

Issue #89, opened 2024-07-31
Proposal for Waveform and Raw HGCROC Hit


ruse-traveler:
![PXL_20240725_235606610](https://github.com/user-attachments/assets/897cc527-f1e5-47db-95d6-ee5451902eac)

One of the items discussed during the [July 2024 ePIC Collaboration Meeting](https://indico.bnl.gov/event/20727/sessions/7437/#20240725) was the steps needed to better simulate the HGCROC. One thing identified was the need to make sure we have the data types necessary to capture both the waveform produced by a SiPM as well as the output of the HGCROC.

**Describe the solution you'd like**

 For the waveform, we could make use of `edm4hep::RawTimeSeries` (from [here](https://github.com/key4hep/EDM4hep/blob/a173bdec749abf5447810b25286a64456914faa6/edm4hep.yaml#L510)).:

```
  edm4hep::RawTimeSeries:
    Description: "Raw data of a detector readout"
    Author: "EDM4hep authors"
    Members:
       - uint64_t cellID  // detector specific cell id
       - int32_t quality               // quality flag for the hit
       - float time [ns]               // time of the hit
       - float charge [fC]             // integrated charge of the hit
       - float interval [ns]           // interval of each sampling
    VectorMembers:
       - int32_t adcCounts          // raw data (32-bit) word at i
```

But for the HGCROC output, we would need to extend what's already in `edm4hep::RawCalorimeterHit`. For example:

```
edm4eic::RawHGCROCHit:
  Description: "Raw hit from an HGCROC"
  Author: "H. G. Croc"
  Members:
    - uint64_t cellID // detector specific (geometrical) cell id
    - int32_t timeOfArrival // ToA value [adc counts]
    - int32_t timeOverThreshold // ToT value [adc counts]
    - int32_t amplitude // amplitude of hit [adc counts]
    - int32_t timeSamp // time stamp for the hit
```

**Describe alternatives you've considered**

One alternate approach could be to save only one of the values from the HGCROC (e.g. amplitude) in `edm4hep::RawCalorimeterHit`

veprbl:
ToA is already part of the edm4hep::RawCalorimeterHit. ToT we could add in our own extension in edm4eic:: namespace, or even submit that upstream. A different question is how to store waveforms between the simulation-type algorithms.

ruse-traveler:
> ToA is already part of the edm4hep::RawCalorimeterHit

Is it? I'm seeing an amplitude field, but not a ToA field...

```
[...]

PR #101, opened 2025-02-09
Add type for HGC/CALOROC output

ruse-traveler:

This PR introduces the `RawCALOROCHit` type, which accommodates the output of both type 1A/1B HGC/CALOROC chips. 

- [ ] Bug fix (issue #__)
- [x] New feature (issue #89)
- [ ] Documentation update
- [ ] Other: __

No.

No.


ruse-traveler:
Leaving this as draft for now while discussion progresses.

One thing that was discussed in [89](https://github.com/eic/EDM4eic/issues/89#issuecomment-2260870531), was to potentially add ToT (and ToA if needed) to `RawCalorimeterHit`. I don't have strong opinions, but considering that systems other than calorimeters (e.g. the PFRICH) will be using HGCROCs, my preference leans towards this being a distinct type... Perhaps another way of framing it is a `RawSiPMHit`?

sly2j:
Nice work! I have a couple of questions/comments:

 - The way we will likely configure the H2GCROC/CALOROC boards for calorimeters would have 5 amplitude reads for each hit so we can do some waveform template fitting. The first read would be before the ToT (to sample the pedestal), the other 4 are during the pulse. Maybe we should consider using something like std::array<int32_t, 5> for amplitude. If we think hard-coding the 5 samples is safe at this point
 - I would personally call the structure after the future EIC ship which would be CALOROC. In particular because there are so many other HGCROC variants out there. The technical name for the current version would be H2GCROC3 for the SiPM version.
 - How do you suggest we deal with storing the phase of the ADC samples? Due to the nature of the 100MHz bunch crossing rate versus the 40MHz clock on the CALOROC board there is a bit of a drift in the timing offset w.r.t. the EIC clock depending on where we are in the 40MHz cycle (both clocks sync up every 50 ns and after that the ASIC sampling drifts by 10ns until they are back in phase). Should we store the phase in a dedicated field? I believe the ASIC (or FEB) can add the phase to the data stream?

@wdconinc @veprbl what do you think? I believe it's important to get this right from the start as changes to the model are expensive (invalidate old files).

ruse-traveler:
Thanks for the feedback @sly2j ! I've in-lined a few thoughts below...

> * The way we will likely configure the H2GCROC/CALOROC boards for calorimeters would have 5 amplitude reads for each hit so we can do some waveform template fitting. The first read would be before the ToT (to sample the pedestal), the other 4 are during the pulse. Maybe we should consider using something like std::array<int32_t, 5> for amplitude. If we think hard-coding the 5 samples is safe at this point

Since they will be configurable (@novitzky also pointed this out to me offline), I wonder if we could go ahead and make the ToT a std::vector<int32_t>?

> * I would personally call the structure after the future EIC ship which would be CALOROC. In particular because there are so many other HGCROC variants out there. The technical name for the current version would be H2GCROC3 for the SiPM version.

Gotcha! Makes sense!

> * How do you suggest we deal with storing the phase of the ADC samples? Due to the nature of the 100MHz bunch crossing rate versus the 40MHz clock on the CALOROC board there is a bit of a drift in the timing offset w.r.t. the EIC clock depending on where we are in the 40MHz cycle (both clocks sync up every 50 ns and after that the ASIC sampling drifts by 10ns until they are back in phase). Should we store the phase in a dedicated field? I believe the ASIC (or FEB) can add the phase to the data stream?

Good question! 😅 I'd definitely be in favor of adding a field for the phase! I feel like that would be the easiest approach for book keeping...

veprbl:
Is the clock counter available in *ROC's?

Also, I think, it would do us good if this structure could also be used to support ongoing beam tests data-taking. cc @tlprotzman

simonge:
What is the use case for storing the phase for every hit? This should always be easily calculatable from the global time of arrival and some calibration offset for the detector at an event level.

Here, is the time stamp the 40MHz clock cycle being added by the FEB and the Time of Arrival from the ASIC in the cycle digitized to 25ps? Can these be combined at this point or might there be further calibrations dependent on having a separate Time of Arrival?

veprbl:
> What is the use case for storing the phase for every hit? This should always be easily calculatable from the global time of arrival and some calibration offset for the detector at an event level.

Would be nice to check if there is any way and desire to synchronize the phases between different chips. Storing the same value should not be too expensive in columnar format with compression.

simonge:
> 2. To account for double hits (e.g. from background), it could be useful to upgrade the ToA and ToT to vector members.
> * Towards (2), the ToA and ToT values have been upgraded to vector members
This will probably be my ignorance of how the CALOROC works, I'm not sure I understand how a single hit will have multiple ToA and ToT values. Could you point me to some documentation/presentation.

> Lastly, I have naïve question for everyone: should `timeStamp` be an int32_t -- in units of TDC(??) -- or a float -- in units of ns?
timeStamp will need to be the digitized value, presumably from the driving 40MHz clock.

edm4eic.yaml
```diff
@@ -244,6 +244,27 @@ datatypes:
...
+      - int32_t  type              // Chip type, 0 - type 1A, 1 - type 1B
+      - uint64_t cellID            // Detector specific (geometrical) cell id
+      - int32_t  samplePhase       // Phase of samples in [# samples], for synchronizing across chips
+      - int32_t  timeStamp         // [TDC]
+    VectorMembers:
+      - int32_t  amplitude         // Iff type == 0, amplitudes of waveforms are stored here [adc counts]
+      - int32_t  lowGainAmplitude  // Iff type == 1, low gain amplitudes of waveforms are stored here [adc counts]
+      - int32_t  highGainAmplitude // Iff type == 1, high gain amplitudes of waveforms are stored here [ac counts]]
+      - int32_t  timeOfArrival     // Calculated times of arrival [TDC]
+      - int32_t  timeOfThreshold   // Calculated times over threshold [TDC]
```
ruse-traveler, line 261:
```suggestion
      - int32_t  timeOverThreshold   // Calculated times over threshold [TDC counts]
```

edm4eic.yaml
```diff
@@ -244,6 +244,27 @@ datatypes:
...
+    Members:
+      - int32_t  type              // Chip type, 0 - type 1A, 1 - type 1B
+      - uint64_t cellID            // Detector specific (geometrical) cell id
+      - int32_t  samplePhase       // Phase of samples in [# samples], for synchronizing across chips
+      - int32_t  timeStamp         // [TDC]
+    VectorMembers:
+      - int32_t  amplitude         // Iff type == 0, amplitudes of waveforms are stored here [adc counts]
+      - int32_t  lowGainAmplitude  // Iff type == 1, low gain amplitudes of waveforms are stored here [adc counts]
+      - int32_t  highGainAmplitude // Iff type == 1, high gain amplitudes of waveforms are stored here [ac counts]]
+      - int32_t  timeOfArrival     // Calculated times of arrival [TDC]
```
ruse-traveler, line 260:
```suggestion
      - int32_t  timeOfArrival     // Calculated times of arrival [TDC counts]
```

edm4eic.yaml
```diff
@@ -244,6 +244,27 @@ datatypes:
...
   ## ==========================================================================
+
+  edm4eic::RawCALOROCHit:
+    Description: "Raw hit from a CALOROC/HGCROC chip"
+    Author: "D. Anderson, S. Joosten, N. Novitzky"
+    Members:
+      - int32_t  type              // Chip type, 0 - type 1A, 1 - type 1B
+      - uint64_t cellID            // Detector specific (geometrical) cell id
+      - int32_t  samplePhase       // Phase of samples in [# samples], for synchronizing across chips
+      - int32_t  timeStamp         // [TDC]
```
ruse-traveler, line 255:
```suggestion
      - int32_t  timeStamp         // [TDC counts]
```

edm4eic.yaml
```diff
@@ -244,6 +244,27 @@ datatypes:
...
+    VectorMembers:
+      - int32_t  amplitude         // Iff type == 0, amplitudes of waveforms are stored here [adc counts]
+      - int32_t  lowGainAmplitude  // Iff type == 1, low gain amplitudes of waveforms are stored here [adc counts]
+      - int32_t  highGainAmplitude // Iff type == 1, high gain amplitudes of waveforms are stored here [ac counts]]
```
ruse-traveler, line 259:
```suggestion
      - int32_t  amplitude         // Iff type == 0, amplitudes of waveforms are stored here [ADC counts]
      - int32_t  lowGainAmplitude  // Iff type == 1, low gain amplitudes of waveforms are stored here [ADC counts]
      - int32_t  highGainAmplitude // Iff type == 1, high gain amplitudes of waveforms are stored here [ADC counts]
```

ruse-traveler (commented):
Fix typos and make capitalizations/units consistent

ruse-traveler:
> This will probably be my ignorance of how the CALOROC works, I'm not sure I understand how a single hit will have multiple ToA and ToT values. Could you point me to some documentation/presentation.

Apologies for the delayed response! 😵 

But, unfortunately, I've been having a tough time finding some documentation on how the multiple ToT/ToA will work in practice (e.g. in test beam or running conditions)... @novitzky do you know of any good documentation on this topic?

> timeStamp will need to be the digitized value, presumably from the driving 40MHz clock.

Gotcha! Thanks! I'll leave the units as [TDC counts], then!

edm4eic.yaml
wdconinc, line 1: Having just fixed a whole bunch of signed/unsigned comparisons, can we please use unsigned ints where appropriate?

ruse-traveler: Good point! Most of these can be `uint32_t`s...

edm4eic.yaml
```diff
@@ -244,6 +244,27 @@ datatypes:
...
+    Author: "D. Anderson, S. Joosten, N. Novitzky"
+    Members:
+      - uint32_t type              // Chip type, 0 - type 1A, 1 - type 1B
+      - uint64_t cellID            // Detector specific (geometrical) cell id
+      - int32_t  samplePhase       // Phase of samples in [# samples], for synchronizing across chips
+      - int32_t  timeStamp         // [TDC counts]
+    VectorMembers:
+      - uint32_t amplitude         // Iff type == 0, amplitudes of waveforms are stored here [ADC counts]
+      - uint32_t lowGainAmplitude  // Iff type == 1, low gain amplitudes of waveforms are stored here [ADC counts]
+      - uint32_t highGainAmplitude // Iff type == 1, high gain amplitudes of waveforms are stored here [ADC counts]
```
veprbl, line 259:
```suggestion
      - uint32_t amplitude         // Iff type == 0, amplitudes of waveform are stored here [ADC counts]
      - uint32_t lowGainAmplitude  // Iff type == 1, low gain amplitudes of waveform are stored here [ADC counts]
      - uint32_t highGainAmplitude // Iff type == 1, high gain amplitudes of waveform are stored here [ADC counts]
```
Only one waveform is stored? (not necessarily a single "pulse") If not, how do we distinguish them.

ruse-traveler: Oh! Good catch! That's correct: it should be only 1 waveform...

PR #106, opened 2025-04-01
Add `SimPulse` data type for storing simulated pulses pre-digitization

sly2j:

This PR adds a new data type, `edm4eic::SimPulse`, to represent simulated pulses before digitization.

The new structure is modeled consistently with the edmhep simulated hit types and aligns with existing waveform structures such as `edm4hep::RawTimeSeries` (generic digitization output) and `edm4hep::TimeSeries` (generic measured time series).

`SimPulse` defines four one-to-many relations relations:
- **`SimCalorimeterHit`**: Pulses can be constructed from contributions of one or more simulated calorimeter hits.
- **`SimTrackerHit`**: Pulses can be constructed from one or more simulated tracker hits.
- **`SimPulse`**: Pulses can be constructed out of other pulses.
- **`MCParticle`**: Each pulse can be attributed to a single MCParticle (primary pulse), multiple MCParticles (overlayed pulses), or zero MCParticles (noise pulse).

Storing pulses at the simulation level ensures conceptual clarity and consistency within our data model. The simulated pulse data to be stored in this structure have no equivalent in real detector output, and are therefore not digitization output.

By using a Simulation structure, we gain direct `Relation` support with other simulation entities, enabling efficient backward navigation through the simulation chain. While Associations are used to link digitization and reconstruction structures to simulation, Relations are preferred for internal connections within the same domain (e.g. within Simulation or within Reconstruction). This design avoids introducing redundant Association types and keeps the data model consistent, clean and efficient.

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [X] Documentation has been added / updated
- [X] Changes have been communicated to collaborators

No

No


ruse-traveler (commented):
Found a typo!

edm4eic.yaml
```diff
@@ -244,6 +244,22 @@ datatypes:
   ## ==========================================================================
   ## Calorimetry
   ## ==========================================================================
+  edm4eic::SimCalorimeterPulse:
+    Description: "Simulated calorimeter pulse prior to digitization."
+    Author: "D. Anderson, S. Joosten"
+    Members:
+      - uint64_t            cellID          // ID of the readout cell for this pulse.
+      - float               energy          // Total energy for this pulse in [GeV].
+      - float               position        // Position of the hit in world coordinates [mm].
```
veprbl, line 253: (comment from @simonge and @ruse-traveler )
Should be a Vector3f.

edm4eic.yaml
```diff
@@ -244,6 +244,22 @@ datatypes:
   ## ==========================================================================
   ## Calorimetry
   ## ==========================================================================
+  edm4eic::SimCalorimeterPulse:
```
veprbl, line 247: We would really want to have generic structures to be able to consistently handle basic operations (adding noise, merging pulses, etc) for different detectors, not just calorimeters.

Would it be acceptable to make a more generic-sounding name, even if we keep the edm4hep::SimCalorimeterHit relation?

edm4eic.yaml
```diff
@@ -244,6 +244,22 @@ datatypes:
...
+      - float               energy          // Total energy for this pulse in [GeV].
+      - float               position        // Position of the hit in world coordinates [mm].
+      - float               time            // Start time for the pulse in [ns].
+      - float               interval        // Time interval between amplitude values [ns].
+    VectorMembers:
+      - float               amplitude       // Pulse amplitude in [GeV], sum of amplitude values equals total energy
+    OneToManyRelations:
+      - edm4hep::SimCalorimeterHit hits     // SimCalorimeterHits used to create this pulse
+      - edm4hep::SimCalorimeterPulse pulses // SimCalorimeterPulses used to create this pulse
+      - edm4hep::MCParticle MCParticle      // MCParticle that caused the pulse
```
veprbl, line 261:
```suggestion
      - edm4hep::MCParticle particle        // MCParticle that caused the pulse
```

edm4eic.yaml
wdconinc: See also a similar change pushed through in edm4hep for technical reasons outlined there; https://github.com/key4hep/EDM4hep/commit/cb738b0e2a5a8d4c922aa2a0d49974caf4adcd2e#diff-41291f568217a9755476ae21a12825127ae5a00179b19e867cffd64e3ceea36f

edm4eic.yaml
wdconinc: See also a similar change pushed through in edm4hep for technical reasons outlined there; https://github.com/key4hep/EDM4hep/commit/cb738b0e2a5a8d4c922aa2a0d49974caf4adcd2e#diff-41291f568217a9755476ae21a12825127ae5a00179b19e867cffd64e3ceea36f

edm4eic.yaml
```diff
@@ -244,6 +244,22 @@ datatypes:
...
+    Description: "Simulated calorimeter pulse prior to digitization."
+    Author: "D. Anderson, S. Joosten"
+    Members:
+      - uint64_t            cellID          // ID of the readout cell for this pulse.
+      - float               energy          // Total energy for this pulse in [GeV].
+      - float               position        // Position of the hit in world coordinates [mm].
+      - float               time            // Start time for the pulse in [ns].
+      - float               interval        // Time interval between amplitude values [ns].
+    VectorMembers:
+      - float               amplitude       // Pulse amplitude in [GeV], sum of amplitude values equals total energy
```
ruse-traveler, line 257: Comment from @veprbl during the meeting: should the units here be GeV? Or should these be in units of voltage?

If we choose to generalize the structure, then we should revisit this. Maybe it could be reformulated to be more generic; perhaps similar to the [edm4hep::TimeSeries](https://github.com/key4hep/EDM4hep/blob/cea6f650aaddc52ebae65ca80c09c18bec168a76/edm4hep.yaml#L593)...

edm4eic.yaml
ruse-traveler, line 253: Comment from @simonge during meeting: the energy and position can technically be obtained from the amplitude vector and cell ID respectively.

veprbl: ... or from the related hit.

edm4eic.yaml
wdconinc: But this is a truth type, not a digitized type. There is no concept of voltage until you have a digitization model. You could argue that energy is not the right quantity for a photon pulse traveling down a fiber either, but it probably should be closer to energy than to voltage.

veprbl: Measured energy deposition is not "truth", it's just another detail of detector technology. Also, there is not much practical use of truth-only type. You will want to process those pulses, and the type of amplitude scale will change along the way, the units won't stay as "GeV".

wdconinc: It's truth in the sense that it is exactly what Geant4 gives. It is like energy of calo hit contributions. Yes, you can change the size of the volume those go in, but it is the exact energy contribution for the tracks in that volume, without distorting effects such as poisson distributions, gaussian measurement uncertainty, or bit depth quantization. It is so "true" that in many cases it is not even representative because it is just one sample out of many possibilities.

edm4eic.yaml
sly2j: Fixed

edm4eic.yaml
sly2j: I added a SimPulse interface type that can be used if this becomes needed in the future (then a SimTrackerPulse can also implement the SimPulse interface).

edm4eic.yaml
sly2j: I changed the name from `energy` to `integral` consistent with `amplitude`. I removed the reference to GeV in the documentation as depending on the system this could be energy or for example a function of the number of photons. I changed the documentation to amplitude to read that **summing** all amplitudes gives the integral (to clarify they don't need to be multiplied by sample width, i.e. they are not differential). This is a choice and can be discussed if we think this is the right one.

edm4eic.yaml
```diff
@@ -579,3 +595,18 @@ datatypes:
     OneToOneRelations:
       - edm4eic::Cluster  cluster           // reference to the cluster
       - edm4eic::Track track                // reference to the track
+
+interfaces:
+  edm4eic:SimHit:
```
veprbl, line 600: This is not a solution. There is no framework support for interfaces, and even if there was, they can't be outputs of an algorithm (e.g. because they do not allow for mutable access).

edm4eic.yaml
veprbl: Still says GeV on this line.

sly2j: Fixed

edm4eic.yaml
sly2j: I disagree. It does provide a solution for all cases except writing for obvious reasons. If needed, algorithms can query the type of the underlying structure and deal with it accordingly.  

Our four options I see are:
  - Use interfaces, which is the canonical way to proceed in podio
  - Don't do anything and have some level of code duplication. This honestly may be fine at this point, but of course we don't like code duplication :-).
  - Use templates, and instantiate them for the correct types to at avoid code duplication. This could likely be automized in future C++ standards using compile-time introspection. Note that this could still use the interface type under the hood, treating it more like a concept.
  - Try to be as generic as we can imagine being. I think that is a design mistake that will ultimately make the code unnecessarily clunky and obtuse.
  
For better or worse we inherit 2 simulation hit types from edm4hep: `SimTrackerHit` and `SimCalorimeterHit`. I believe that sticking with this structure would provide the most consistent and predictable structure for the end users.

Framework support not being there is irrelevant IMO - if we need it we would add it. The discussion is really at a higher level: Do we decide to define interface types or not in our data model. Once the interface is there we can worry about implementation and guidance for users on how we want them used.

I don't mind punting on this part and spinning it into a separate PR as I don't want to hold up `SimCalorimeterPulse` from being available for algorithm development. 

**What does everyone think? Should we briefly meet today or Monday to chat and decide on the path forward so we can keep moving?**

veprbl: The solution is outlined in https://github.com/eic/EDM4eic/pull/106#discussion_r2035793944 .

edm4eic.yaml
simonge: Just to clear up my understanding, the hit position is the interaction position on in the crystal/fibre while the pulse position will be the position of the SiPM?

edm4eic.yaml
sly2j: You are suggesting to change the name without changing the structure. Philosophically, I don't think that's the right way to go, but it's something we can discuss. At the other hand, there's in fact a precedent where we use "Cluster" instead of "CalorimeterCluster" in the reconstruction. 

However, I could see the need for an analogous but somewhat different structure for e.g larger MCP-PMT hits, which use `SimTrackerHit` (because we use `SimTrackerHit`s for optical photons). Anyway, if we're just discussing whether or not to change the name from `SimCalorimeterHit` to, e.g. `SimulatedPulse` or `SimPulse` we can discuss this on Monday during the Reconstruction meeting. 

Ultimately, most of all I don't want to delay actual work bickering over this. We're starving for FTEs, so we should capitalize on what we have now as it could disappear tomorrow... And it's ultimately not _that_ expensive to change a name in the data model - we've done it before :-).

edm4eic.yaml
```diff
@@ -244,6 +244,22 @@ datatypes:
   ## ==========================================================================
   ## Calorimetry
   ## ==========================================================================
+  edm4eic::SimCalorimeterPulse:
+    Description: "Simulated calorimeter pulse prior to digitization, implements SimPulse."
```
veprbl, line 248:
```suggestion
    Description: "Simulated calorimeter pulse prior to digitization, implements SimPulse"
```

edm4eic.yaml
```diff
@@ -244,6 +244,22 @@ datatypes:
...
+      - float               integral        // Total pulse integral in relevant units.
+      - edm4hep::Vector3f   position        // Position of the hit in world coordinates [mm].
+      - float               time            // Start time for the pulse in [ns].
+      - float               interval        // Time interval between amplitude values [ns].
+    VectorMembers:
+      - float               amplitude       // Pulse amplitude in in relevant units, sum of amplitude values equals integral
+    OneToManyRelations:
+      - edm4hep::SimCalorimeterHit hits     // SimCalorimeterHits used to create this pulse
+      - edm4eic::SimCalorimeterPulse pulses // SimCalorimeterPulses used to create this pulse
+      - edm4hep::MCParticle particle        // MCParticle that caused the pulse
```
veprbl, line 261:
```suggestion
      - edm4hep::MCParticle particles       // MCParticle that caused the pulse
```
All one to many relations are plural, it looks like.

edm4eic.yaml
veprbl:
```suggestion
    Description: "Simulated calorimeter pulse prior to digitization"
```
Once we remove SimPulse

edm4eic.yaml
```diff
@@ -203,6 +203,27 @@ datatypes:
...
+  ## Simulation info
+  ## ==========================================================================
+
+  edm4eic::SimPulse:
+    Description: "Simulated pulse prior to digitization."
+    Author: "D. Anderson, S. Gardner, S. Joosten., D. Kalinkin"
+    Members:
+      - uint64_t            cellID          // ID of the readout cell for this pulse.
+      - float               integral        // Total pulse integral in relevant units.
+      - edm4hep::Vector3f   position        // Position of the hit in world coordinates [mm].
```
simonge, line 216:
```suggestion
      + edm4hep::Vector3f   position        // Position the pulse is evaluated in world coordinates [mm].
```
Or something similar to add clarity.

edm4eic.yaml
```diff
@@ -203,6 +203,27 @@ datatypes:
...
+  ## Simulation info
+  ## ==========================================================================
+
+  edm4eic::SimPulse:
+    Description: "Simulated pulse prior to digitization."
+    Author: "D. Anderson, S. Gardner, S. Joosten., D. Kalinkin"
+    Members:
+      - uint64_t            cellID          // ID of the readout cell for this pulse.
+      - float               integral        // Total pulse integral in relevant units.
+      + edm4hep::Vector3f   position        // Position the pulse is evaluated in world coordinates [mm].
```
veprbl, line 216:
```suggestion
      - edm4hep::Vector3f   position        // Position the pulse is evaluated in world coordinates [mm].
```
Unless this is correct yaml

veprbl (commented):
Could you also update version in CMakeLists.txt, please?

edm4eic.yaml
```diff
@@ -203,6 +203,27 @@ datatypes:
...
+      - float               time            // Start time for the pulse in [ns].
+      - float               interval        // Time interval between amplitude values [ns].
+    VectorMembers:
+      - float               amplitude       // Pulse amplitude in in relevant units, sum of amplitude values equals integral
```
simonge, line 220:
```suggestion
      - float               amplitude       // Pulse amplitude in relevant units, sum of amplitude values equals integral
```

veprbl (approved):

PR #116, opened 2025-06-09
An alternative implementation of a type for HGCROC output

ruse-traveler:

This PR introduces two types to facilitate work with the output of HGCROC chips:
1. The component, `edm4eic::HGCROCSample`, which defines for a given sample read out by the a chip
    - the ADC,
    - the Time of Arrival (TOA),
    - the Time over Threshold (TOT),
    - and the in-progress/complete relevant flags for TOT;
2. And the datatype, `edm4eic::RawHGCROCHit`, which consolidates for a given hit
    - the geometrical information (cell ID)
    - the time information (time stamp and sample phase),
    - and the relevant samples organized into a time series (represented as a vector member).

This PR is intended to supersede #101. It makes two distinct design choices from #101 informed by experience from the 2024 LFHCAL and EEEMCal test beams, both of which utilized HGCROC prototypes.
1. Rather than define separate vector members for the ADC, TOA, and TOT values, these values are collected into a component which are then stored as a vector. This more closely matches what the HGCROC actually reads out, where each value is technically read out once per sample (even if some, such as the TOA and TOT are zero for most).
2. The dual readout ("Type B") chips are intentionally **not** supported by this type. This is because such chips are currently not available nor in use. It is assumed that the processing of such a chip would be significantly different than the current chips actually in use ("Type A"), and as such would be better supported by a distinct type.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #89 )
- [ ] Documentation update
- [ ] Other: __

No.

No.


edm4eic.yaml
```diff
@@ -191,6 +191,15 @@ components:
       - uint32_t boundValuesSize          // size of bound values
       - std::array<double,16> transform   // row-wise 4x4 affine transform [R T; 0 1] with 3x3 rotation matrix R and translation column 3-vector T
 
+  ## An individual sample output by a HGC/CALOROC chip
+  edm4eic::CALOROCSample:
```
veprbl, line 195:
```suggestion
  ## An individual sample output by a HGCROC chip
  edm4eic::HGCROCSample:
```
As discussed, we don't know much about CALOROC, so this is HGCROC

edm4eic.yaml
```diff
@@ -265,6 +274,17 @@ datatypes:
   ## ==========================================================================
   ## Calorimetry
   ## ==========================================================================
+
+  edm4eic::RawCALOROCHit:
+    Description: "Raw hit from a CALOROC/HGCROC chip"
```
veprbl, line 279:
```suggestion
  edm4eic::RawHGCROCHit:
    Description: "Raw hit from a HGCROC chip"
```

edm4eic.yaml
```diff
@@ -265,6 +274,17 @@ datatypes:
...
+    Author: "D. Anderson, S. Joosten, T. Protzman, N. Novitzky"
+    Members:
+      - uint64_t cellID                // Detector specific (geometrical) cell id
+      - int32_t  samplePhase           // Phase of samples in [# samples], for synchronizing across chips
+      - int32_t  timeStamp             // [TDC counts]
+    VectorMembers:
+      - edm4eic::CALOROCSample sample  // ADC, Time of Arrival (TOA), and Time over Threshold (TOT) values for each sample read out
```
veprbl, line 286:
```suggestion
      - edm4eic::HGCROCSample sample  // ADC, Time of Arrival (TOA), and Time over Threshold (TOT) values for each sample read out
```

veprbl:
Thank you for addressing my comments, @ruse-traveler . This looks good to me. I have no further suggestions.

tlprotzman:
This looks good to me as well.  The only thing I'm not totally confident about is how the phase will actually be used, but I agree it is good to keep in the data type.

veprbl (dismissed):

ruse-traveler:
Sorry for the review re-request @veprbl , but I just caught that the sample vector wasn't pluralized in the raw hit 😓

veprbl (approved):

PR #124, opened 2025-09-21
Remove irrelevant comment for edm4eic::ReconstructedParticle::PDG

veprbl:
The PDG field was always there:
https://github.com/key4hep/EDM4hep/blob/f5ee28748cad196b794fd7c84af2d1660dce3684/edm4hep.yaml#L557 There is no deviation, if we don't like the field we should just remove it.


ruse-traveler (approved):
Good catch! I'm okay with removing this.

PR #128, opened 2025-10-29
feat: add Truthiness datatype for Monte Carlo comparisons

wdconinc:
This PR adds a datatype to record the "[truthiness](https://en.wikipedia.org/wiki/Truthiness)" (as mathematically defined...) for a reconstructed event; where truthiness is the "quality of seeming or being felt to be true, even if not necessarily true," or in this case also "the amount of confidently proclaiming the wrong thing to be true."

Mathematically, truthiness is a non-negative value that is zero only for perfectly reconstructed events (positive-definite), and is radially increasing in the error of the reconstruction (greater error leads to greater truthiness).

It is possible to define truthiness in multiple ways, but we will typically use some combination of the following components:
- a &chi;<sup>2</sup> measure on associated reconstructed and generated particles, with normalization given by the determined uncertainty in the reconstruction (if available) or 1 GeV otherwise,
- a positive penalty term for discrete reconstruction errors, such as PID mis-identification (where weighting can be used to penalize some mis-identification more than others),
- a positive penalty term for generated particles that should have been reconstructed, but weren't,
- a positive penalty term for reconstructed particles that were not part of the original event record.

There are non-reconstruction reasons why the truthiness will be non-zero in realistic scenarios:
- multiple-scattering effects will cause the event to lose momentum starting from the true value, deviating both in direction and magnitude in a consistent direction,
- secondary particles will be generated in materials or along bent trajectories, leading to additional reconstructed particles corresponding to e.g. hard bremsstrahlung gammas in the electromagnetic calorimeters,
- primary particles (in particular are low energies) may be absorbed in support structures, leading to their absence in the reconstructed event.

Nevertheless, the decrease of the overall average event truthiness for the same geometry and input hit collections is intended to indicate an improved reconstruction, and converse.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: store truthiness for event reconstruction)
- [ ] Documentation update
- [ ] Other: __

No.

No.


edm4eic.yaml
```diff
@@ -614,6 +614,32 @@ datatypes:
...
+                  with non-negative contributions from various aspects of the reconstruction,
+                  where a zero value indicates a perfect reconstruction."
+    Author: "W. Deconinck, S. Colbert"
+    Members:
+       - float event_truthiness                       // Overall truthiness of the entire event
+       - float event_pid_contribution                 // Contribution of PID matching to truthiness
+       - float event_energy_contribution              // Contribution of energy matching to truthiness
+       - float event_momentum_contribution            // Contribution of momentum matching to truthiness
+       - float unassociated_mc_particles_contribution // Contribution from unassociated MC particles
+       - float unassociated_rc_particles_contribution // Contribution from unassociated reconstructed particles
```
ruse-traveler: For consistency with other types, I'd suggest using camel case for field names.
```suggestion
       - float eventTruthiness                       // Overall truthiness of the entire event
       - float eventPIDContribution                 // Contribution of PID matching to truthiness
       - float eventEnergyContribution              // Contribution of energy matching to truthiness
       - float eventMomentumContribution            // Contribution of momentum matching to truthiness
       - float unassociatedMCParticlesContribution // Contribution from unassociated MC particles
       - float unassociatedRecoParticlesContribution // Contribution from unassociated reconstructed particles
```

ruse-traveler (commented):
Only comments I have so far are the component-idea for consideration above, and some suggestions for naming consistency between this and other types' fields below!

ruse-traveler (approved):
🙌 Thanks! I think this looks good!

Issue #127, opened 2025-10-28
Proposal for a Secondary Vertex Type


ruse-traveler:
As secondary vertexing work advances on both helix swimming techniques and the ACTS AMVF, it has become clear that additional types may be needed to encode information related to secondaries and resonance decays. This will be particularly critical for
1. benchmarking secondary vertexing algorithms against each other, and
2. facilitating comparisons between algorithms implemented in EICrecon vs. those ran in standalone analyses.

**Describe the solution you'd like**

A proposal was advanced by @starsdong in the [10.20.2025 Reco WG meeting](https://indico.bnl.gov/event/30168):

```yaml
edm4eic::SecondaryVertex:
  Description: "EIC secondary vertex"
  Author: "X. Dong"
  Members:
    - int32_t             type          // Type flag, to identify what type of vertex it is (e.g. primary, secondary, generated, etc.)
    - float               chi2          // Chi-squared of the vertex fit
    - int                 ndf           // NDF of the vertex fit
    - edm4hep::Vector4f   position      // position [mm] + time t0 [ns] of the vertex. Time is the 4th component in vector
    - edm4eic::Cov4f      positionError // Covariance matrix of the position+time. Time is the 4th component, similarly to 4vector
    - edm4hep::Vector3f   parentMomentum        // parent momentum
    - edm4hep::Vector3f   parentDecayLength     // parent decay length L
    - float               parentInvariantMass   // parent invariant mass
    - float               parentDecayLengthChi2 // parent L/dL
    - float               parentDca2PV          // parent dca to primary vertex
  VectorMembers:
    - edm4hep::Vector3f   daughterMomentum    // daughter track momentum
    - float               daughterMass        // daughter mass
    - float               daughterDca2PV      // daughter dca to primary vertex
    - float               daughterDca2PVChi2  // daughter dca/sigma to primary vertex
    - float               daughterPairIndices // pair track indices
    - float               daughterPairDca     // pair dca to primary vertex
    - float               daughterPairDcaChi2 // pair dca/sigma to primary vertex
  OneToOneRelations:
    - edm4eic::Vertex     primaryVertex // associated primary vertex
  OneToManyRelations:
    - edm4eic::ReconstructedParticle associatedParticles // particles associated to this vertex.
```

**Describe alternatives you've considered**

There are several paths to encoding this information, and are discussed further in the comments.

ruse-traveler:
Responses to me from @starsdong on 10.20.2025:

[...]

PR #132, opened 2025-11-14
added SecondaryVertex container

starsdong:
Add a new container SecondaryVertex to store topological variables used for secondary vertex reconstruction

- [ ] Bug fix (issue #__)
- [x ] New feature (issue #127 )
- [ ] Documentation update
- [ ] Other: __

- [x] Tests for the changes have been added

None

None

### DIFF
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -9,7 +9,7 @@
 ## If there are schema version changes that can be evolved, see the podio documentation
 ## for an example: https://github.com/AIDASoft/podio/tree/master/tests/schema_evolution
 ##
-schema_version: 840
+schema_version: 850
 
 options :
   # should getters / setters be prefixed with get / set?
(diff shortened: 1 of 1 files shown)


edm4eic.yaml
```diff
@@ -511,6 +511,35 @@ datatypes:
...
+      - edm4eic::Cov4f      positionError // Covariance matrix of the position+time. Time is 4th component, similarly to 4vect
+      - edm4hep::Vector3f   parentMomentum             // parent momentum
+      - float               parentInvariantMass        // parent invariant mass         
+      - float               parentInvariantMassError   // parent invariant mass error         
+      - float               parentDecayLength          // parent decay length
+      - float               parentDecayLengthError     // parent decay length error
+      - float               parentDca2PV               // parent dca to primary vertex
+      - float               parentDca2PVError          // parent dca_error to primary vertex
+    VectorMembers:
+      - edm4hep::Vector3f   daughterMomentum           // daughter track momentum at the decay vertex
```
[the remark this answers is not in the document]
starsdong: Wouter, I agree that a full covariance matrix including momentum with positions (DCA positions of the tracks w.r.t the reconstructed vertex) would be more complete. From previous experience to my knowledge, the relevant covariance matrices were used in reconstruction and fitting, but not really used in the downstream analysis. Most of the downstream analysis will work on optimization with the selection cuts on various topology (error-normalized) variables saved here. But if there is an interest to keep them, it shouldn't be difficult to add the full covariance in.

edm4eic.yaml
```diff
@@ -511,6 +511,35 @@ datatypes:
...
+      - int                 daughterPDG                // daughter PDG
+      - float               daughterDca2PV             // daughter dca to primary vertex
+      - float               daughterDca2PVError        // daughter dca_error to primary vertex
+      - int                 daughterPairIndices        // track indices for any pair
+      - float               daughterPairDca            // dca between any pair of tracks
+      - float               daughterPairDcaError       // dca_error between any pair of tracks
+    OneToOneRelations:
+      - edm4eic::Vertex     primaryVertex              // associated primary vertex
```
ruse-traveler: For my own understanding: is there a reason why it we would always want to link to only the primary vertex?

I'm imagining something like an upsilon decay where we might have an intermediate resonance that decays further, e.g. an upsilon. In which case, the relevant decay length for the intermediary (like a D0) would be between the upsilon vertex and the intermediary's decay vertex...

edm4eic.yaml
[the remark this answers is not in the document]
ruse-traveler: I think adding that info is the way to go! I understand that right now that only the 3-momenta is really used in analysis, but I think that offering a more complete set of information on the daughters goes a long way towards ensuring that the datatype (or datatypes) implemented here can support a much wider array of analyses in the near(ish) term.

edm4eic.yaml
Simple-Shyam: Hi, yes the helix method is used for the current analysis. It allows to extract all the information. The only drawback is that it ignores the errors as it's an analytical approach.

wdconinc: > it ignores the errors as it's an analytical approach

(OT) The former does not follow from the latter. The helix method just doesn't propagate errors, and that's a choice.

Simple-Shyam: Let me say again, it can calculate DCA, and all other topological variables but not the errors because this method doesn't consider the covariance.

Issue #133, opened 2025-11-22
Add CALOROC type B data structure


mhkim-anl:
This issue is to add CALOROC type B data structure to study its measurement performance for various calorimeters.

PR #134, opened 2025-11-23
Add CALOROC1B data model

mhkim-anl:
This PR introduces a data model for the CALOROC1B chip. Although it is not yet available in hardware, it is, together with the HGCROC chip, a strong candidate for the calorimeter readout in the ePIC experiment. Having a dedicated CALOROC1B data model allows us to study its expected performance in detail. The performance not only can be directly compared with the HGCROC-based performance, but also provide useful input for the CALOROC1B chip development.

Since the main difference between the CALOROC1B and HGCROC (CALOROC1A) chip is whether the chip has TOT or low gain ADC, the data model has been implemented based on `edm4eic::HGCROCSample` and `edm4eic::RawHGCROCHit`.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #133)
- [ ] Documentation update
- [ ] Other: __

No
No

### DIFF
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -195,8 +195,8 @@ components:
   edm4eic::CALOROC1ASample:
     Members:
       - uint16_t ADC                // [ADC Counts], amplitude of signal during sample, valid IFF TOTInProgress is false
-      - bool     TOTInProgress      // Flag which indicates if TOT calculation is ongoing, ADC value may be corrupted if this is true
-      - bool     TOTComplete        // Flag which indicates if a TOT calculation is complete and TOT value is valid
+      - uint16_t timeOfArrival      // Time of arrival (TOA) [TDC counts], nonzero IFF ADC crossed threshold upwards during sample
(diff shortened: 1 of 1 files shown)


ruse-traveler:
> HGCROC is not the same as CALOROC1A?
> If yes, then what is the better name? `RawROCHit`?

Ah! Then I may have misunderstood the discussion... What's the difference between the 1A and the HGC?
I think `RawROCHit`'s a good name if they are different, though!

mhkim-anl:
Thank you very much for arranging the action items and for the discussion on the name of the data model! If I remember correctly, the data structure of the HGCROC chip is as same as the CALOROC1A (I'll cross-check this point). So I've updated the `edm4eic.yml` accordingly. Please let me know if I've misunderstood anything from the last simulation meeting.

mandrenguyen:
Yes, I think the same data structure is appropriate.  The CALOROCA will have a different number of channels and will be adapted for the EIC clock, but it will have the same format (ADC/TOT/TOA).  The CALOROCB will instead by dual-gain ADC (no TOT).

ruse-traveler (approved):
Looks good from my end! Thank you for driving the development on this @mhkim-anl 🥳!

veprbl (commented):
This was approved in the Wednesday meeting.

PR #136, opened 2025-11-28
fix: move units from comments to the podio field for it

wdconinc:
This PR cleans up the edm4eic yaml file a bit:
- move the units from the comments to the field for it in the datamodel syntax,
- use yaml multiline blocks for declarations instead of ugly strings,
- align the variables and comments a bit better.


CMakeLists.txt
```diff
@@ -30,10 +30,14 @@ include(GNUInstallDirs)
...
 include(CTest)
 
-find_package(podio 0.15)
+set(podio_min_version 0.17)
+find_package(podio ${podio_min_version})
 if(NOT podio_FOUND)
   find_package(podio 1.0 REQUIRED)
 endif()
+if(podio_VERSION VERSION_LESS ${podio_min_version} AND podio_VERSION VERSION_LESS "1.0")
+  message(FATAL_ERROR "podio version ${podio_VERSION} is too old. Please use podio >= 0.17")
```
veprbl:
```suggestion
  message(FATAL_ERROR "podio version ${podio_VERSION} is too old. Please use podio >= ${podio_min_version}")
```

PR #142, opened 2026-01-17
feat: update schema version to 8.6.0 as well

veprbl:
Something we should have done in 9b4a7751e9fb6664f9989de21baa91989c85e221


wdconinc:
My approval is no good around here :-(

PR #171, opened 2026-06-08
Split `edm4eic::Cluster::shapeParameters` into individual fields

### LINKED ISSUE #85: Break up `Cluster::clusterShape` into named fields
The `clusterShape` vector member had become a dumping ground for additional values that bypass EDM review procedure https://github.com/eic/EDM4eic/blob/f21b3881034524b312921b286081ddfdbe45cff5/edm4eic.yaml#L292
In addition to that, accessing values by index is confusing to our users. We should migrate away from this approach.

See also https://github.com/eic/EICrecon/pull/1511

veprbl:
The current shapeParameters field is to be deprecated.

### DIFF
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -335,9 +335,8 @@ datatypes:
       - edm4eic::Cov3f    positionError     // Covariance matrix of the position (6 Parameters).
       - float             radius       // Cluster radius [mm].
       - float             dispersion   // Cluster dispersion [mm].
-      - float             widthTheta   // Cluster width in polar (theta) direction [rad].
-      - float             widthPhi     // Cluster width in azimuthal (phi) direction [rad].
-      - std::array<float, 3> principalAxesLengths // Lengths along the cluster's principal axes [mm], sorted in descending order (equivalent to sqrt of eigenvalues of the position covariance, descending).
+      - std::array<float, 3> principalAxesLengthsXYZ      // Lengths along the cluster's principal axes [mm], sorted in descending order (equivalent to sqrt of eigenvalues of the position covariance, descending).
+      - std::array<float, 2> principalAxesLengthsThetaPhi // Lengths along the cluster's principal axes [rad], sorted in descending order.
       - float             intrinsicTheta    // Intrinsic cluster propagation direction polar angle [rad]
       - float             intrinsicPhi      // Intrinsic cluster propagation direction azimuthal angle [rad]
       - edm4eic::Cov2f    intrinsicDirectionError // Error on the intrinsic cluster propagation direction
--- a/edm4eic.yaml
+++ b/edm4eic.yaml
@@ -324,7 +324,6 @@ datatypes:
     Description: "EIC hit cluster, reworked to more closely resemble EDM4hep"
     Author: "W. Armstrong, S. Joosten, C.Peng"
     Members:
-      # main variables
       - int32_t           type              // Flag-word that defines the type of the cluster
       - float             energy            // Reconstructed energy of the cluster [GeV].
       - float             energyError       // Error on the cluster energy [GeV]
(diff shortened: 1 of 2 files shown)


edm4eic.yaml
```diff
@@ -333,11 +333,16 @@ datatypes:
       - uint32_t          nhits             // Number of hits in the cluster.
       - edm4hep::Vector3f position          // Global position of the cluster [mm].
       - edm4eic::Cov3f    positionError     // Covariance matrix of the position (6 Parameters).
+      - float             radius       // Cluster radius [mm].
+      - float             dispersion   // Cluster dispersion [mm].
+      - float             widthTheta   // Cluster width in polar (theta) direction [rad].
+      - float             widthPhi     // Cluster width in azimuthal (phi) direction [rad].
+      - std::array<float, 3> principalAxesLengths // Lengths along the cluster's principal axes [mm], sorted in descending order (equivalent to sqrt of eigenvalues of the position covariance, descending).
       - float             intrinsicTheta    // Intrinsic cluster propagation direction polar angle [rad]
       - float             intrinsicPhi      // Intrinsic cluster propagation direction azimuthal angle [rad]
```
akioogawa: // This is angle of long axis = "theta_sigmaMax" in XY for XY detectors

edm4eic.yaml
```diff
@@ -333,11 +333,16 @@ datatypes:
       - uint32_t          nhits             // Number of hits in the cluster.
       - edm4hep::Vector3f position          // Global position of the cluster [mm].
       - edm4eic::Cov3f    positionError     // Covariance matrix of the position (6 Parameters).
+      - float             radius       // Cluster radius [mm].
+      - float             dispersion   // Cluster dispersion [mm].
+      - float             widthTheta   // Cluster width in polar (theta) direction [rad].
+      - float             widthPhi     // Cluster width in azimuthal (phi) direction [rad].
+      - std::array<float, 3> principalAxesLengths // Lengths along the cluster's principal axes [mm], sorted in descending order (equivalent to sqrt of eigenvalues of the position covariance, descending).
```
akioogawa: // For XY detector with constant Z, these are sigmaMax, SigmaMin and 0.

akioogawa (dismissed):
looks good.

ruse-traveler (dismissed):
Looks great! This is a very welcome change, so thank you for taking the initiative on this @veprbl!

I'm fine with this as-is after incorporating @akioogawa's suggestion, but something to consider: maybe the shape parameters could be made into a component? I could see these as _maybe_ being useful in other contexts such as tracking clusters...

edm4eic.yaml
```diff
@@ -333,11 +333,16 @@ datatypes:
       - uint32_t          nhits             // Number of hits in the cluster.
       - edm4hep::Vector3f position          // Global position of the cluster [mm].
       - edm4eic::Cov3f    positionError     // Covariance matrix of the position (6 Parameters).
+      - float             radius       // Cluster radius [mm].
+      - float             dispersion   // Cluster dispersion [mm].
+      - float             widthTheta   // Cluster width in polar (theta) direction [rad].
```
akioogawa: Before https://github.com/eic/EICrecon/pull/2683, shapeParameters[2] and [3] were sigmaMax^2 and sigmaMin^2 in 2d theta-phi space, but unsorted by value. 

Currently after https://github.com/eic/EICrecon/pull/2683, shapeParameters[2] and [3] are 2d sigmaMax^2 and sigmaMin^2 in 2d theta-phi space, but now in descending order.

Unless people who use these (barrel calorimeter?) wants theta & phi width, I'd suggest 
std::array<float, 2> principalAxesLengthThetaPhi, and line below to be principalAxesLengths3d or principalAxesLengthsXYZ.

If people want theta & phi width, and not long and short axis width, we need to drop running SelfAdjointEigenSolver in CalorimeterClusterShape.cc for cov2. Then can keep widthTheta and widthPhi name here.

And we can keep values as is but then need to modify comment to [mm^2] and [rad^2] units, or take sqrt and leave comment [mm] and [rad] as is.

veprbl: Thank you for this well thought-out proposal. I agree, and this is now implemented.

Regarding the squares, I did not mention in my today's talk, but we agreed that the named variables will be taken with sqrt, so [mm] and [rad] are the units.

edm4eic.yaml
veprbl: I've addressed your suggestion with a different wording. Could you, please, check?

edm4eic.yaml
veprbl: Also addressed.

edm4eic.yaml
akioogawa: Looks good

edm4eic.yaml
akioogawa: OK

ruse-traveler (approved):
LGTM!

PR #175, opened 2026-08-11
cmake: build SIO and Arrow backend support

veprbl:
Based on https://github.com/key4hep/EDM4hep/pull/502
Renamed Arrow -> PodioArrow per pending upstream change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,7 +101,7 @@ endif()
 if(PODIO_ENABLE_ARROW)
   PODIO_ADD_ARROW(edm4eic "${headers}" "${sources}")
   message(STATUS "Building and installing the Arrow component since podio supports it")
-  add_library(EDM4EIC::edm4eicArrow ALIAS edm4eicPodioArrow)
+  add_library(EDM4EIC::edm4eicPodioArrow ALIAS edm4eicPodioArrow)
   list(APPEND EDM4EIC_INSTALL_LIBS edm4eicPodioArrow)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,10 +87,23 @@ PODIO_ADD_ROOT_IO_DICT(edm4eicDict edm4eic "${headers}" src/selection.xml
 
 add_library(edm4eic::edm4eicDict ALIAS edm4eicDict )
 
-list(APPEND EDM4EIC_INSTALL_LIBS edm4eic edm4eicDict)
-
 add_subdirectory(utils)
 add_subdirectory(test)
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -87,11 +87,24 @@ PODIO_ADD_ROOT_IO_DICT(edm4eicDict edm4eic "${headers}" src/selection.xml
...
 add_library(edm4eic::edm4eicDict ALIAS edm4eicDict )
 
-list(APPEND EDM4EIC_INSTALL_LIBS edm4eic edm4eicDict)
-
 add_subdirectory(utils)
 add_subdirectory(test)
 
+list(APPEND EDM4EIC_INSTALL_LIBS edm4eic edm4eicDict)
+
+PODIO_ADD_SIO_IO_BLOCKS(edm4eic "${headers}" "${sources}")
```
wdconinc, line 95: Do we need sio? We don't build podio with sio support, and we don't have sio in the container.

veprbl: No. I'm not aware of any plan to use SIO, but it would be consistent with upstream. I'm happy to not include it, if that's the preference.

CMakeLists.txt
```diff
@@ -87,11 +87,24 @@ PODIO_ADD_ROOT_IO_DICT(edm4eicDict edm4eic "${headers}" src/selection.xml
...
+if(TARGET edm4eicSioBlocks)
+  message(STATUS "Building and installing the SioBlocks since podio supports it")
+  list(APPEND EDM4EIC_INSTALL_LIBS edm4eicSioBlocks)
+endif()
+
+if(PODIO_ENABLE_ARROW)
+  PODIO_ADD_ARROW(edm4eic "${headers}" "${sources}")
+  message(STATUS "Building and installing the Arrow component since podio supports it")
+  add_library(EDM4EIC::edm4eicArrow ALIAS edm4eicPodioArrow)
```
wdconinc:
```suggestion
  add_library(EDM4EIC::edm4eicPodioArrow ALIAS edm4eicPodioArrow)
```

CMakeLists.txt
wdconinc: Is this the full extent of what we'd need to add to support sio? Do we need to make sure we build sio in ci to avoid bitrot?

veprbl: This provides the SioBlocks library. It could stop working for hypothetical users of SIO, sure.

wdconinc (approved):

CMakeLists.txt
wdconinc: This broke the podio CI: https://github.com/AIDASoft/podio/actions/runs/32156321288/job/95774482712?pr=1004

